PR #6, opened 2015-08-17
concurrent safe log package for printing, don't remake buf in gpsSerialReader loop

jpoirier:
...low-hanging fruit, I know. :/


cyoung:
What logging package do you like most, "log"?

jpoirier:
I've not really had the need to use anything other than the native log package, but then my needs have been fairly simple.

PR #27, opened 2015-09-01
improvements for mobile device support of the webUI

bradanlane:
made layout responsive to mobile devices
formatted uptime (this would be better done in the socket code)
added mobile web app icon
added prompt to pin webUI to device homescreen
added shape formatting to red/green indicators for accessibility
updated styles to reflect current design recomendations

suggestion for socket data - add RPi board temperature
/sys/class/thermal/thermal_zone0/temp


cyoung:
Excellent! Good suggestion on the RPi board temperature.

cyoung:
https://github.com/cyoung/stratux/commit/ecd29a8425c5bb8b1088a65d9df6ccd6f50444c2

cyoung:
Getting:

     TypeError: time_parts is null

And "Uptime: --". Just after starting.

bradanlane:
Ok. Let me check that. I was getting the error on the local uptime temporary variable but there must be a condition when uptime is defined but the value is unrecognized.

BrantKyser:
What about formatting the uptime on the server side. It looks like Go provides a way to represent time in about any conceivable format.

cyoung:
Sure, how do we want it formatted?

bradanlane:
Actually I was about to suggest the opposite. Separate data from formatting. Keep the format in the JS code

cyoung:
Well, let's do that then. Ball is in your court :)

Here is a string that breaks the current display: "Uptime":"7.004683139s"

bradanlane:
That sting would indeed break my code. Sorry about that. Working on the code now.

bradanlane:
This will take a little while. I'm going to need to build a Dev environment

PR #28, opened 2015-09-02
Make the SD card read-only

kdknigga:
This seems to work.


image/spindle/wheezy-stage4
```diff
@@ -37,33 +37,41 @@ apt-get install -y tcpdump
...
 mv -f /tmp/sshd_config.in /etc/ssh/sshd_config
+mv -f /tmp/mount_unionfs.in /usr/local/bin/mount_unionfs
 rm -f /usr/share/dbus-1/system-services/fi.epitest.hostap.WPASupplicant.service
 
+chmod 755 /usr/local/bin/mount_unionfs
+
+sed -r 's/(\/dev\/mmcblk[[:alnum:]]+([[:space:]]+[[:graph:]]+){2}[[:space:]]+)([[:graph:]]+)([[:space:]]+.*)/\1ro,\3\4/g' -i /etc/fstab
+
+echo "
+mount_unionfs   /etc            fuse    defaults          0       0
```
haykinson: so this leaves /etc and /var as read-write, allowing modifications? what about the /root path? those of us using their install to update to the latest code will want to be able to build right from there, which means installing golang and updating /root/stratux or some other location to the latest code.

kdknigga: /etc and /var become ramdisks, so anything written is discarded on reboot.

I think you should be able to 'mount -o rw,remount /whatever' to
temporarily enable writing.

bradanlane: Something confuses me - If the webUI has the ability to change the startus.conf, then this will somehow need to be persisted between boots.

kdknigga: OK, so what if we keep the ramdisk overlay for /var and leave everything
else read-write.

There shouldn't be much writing under normal operation outside of /tmp and
/var, so that should take care of a large chunk of the possibility for file
system corruption.

Further, perhaps the web interface could have a shutdown button for extra
safety?

JohnOCFII:
What is the goal of making the SD card read-only?  I don't see a particular issue in the GitHub issues to tie to this PR.

bradanlane:
we may be able to create some sort of web interface to a remote shutdown but no one will use in in normal operations so I'd recommend we work on making the unit idiot-proof.

kdknigga:
Trying to protect against file system corruption due to improper shutdown
(yanking the power).

Apparently some people have reported this problem.

JohnOCFII:
> Trying to protect against file system corruption due to improper shutdown
> (yanking the power).

Got it.  Thanks.

rwilmoth:
Any possibility to hard-wire a small button or switch to the pins that's set to a shutdown -h now or similar? Or have bidirectional comms with the app so that when it stops transmitting for a set amount of time it automatically shuts down?

haykinson:
There's this one: http://mausberry-circuits.myshopify.com/collections/frontpage/products/direct-plug-shutdown-switch-with-rocker

kdknigga:
Also:
https://www.reddit.com/r/stratux/comments/3hw2x8/rpi2_power_off_button/

kdknigga:
> I'd recommend we work on making the unit idiot-proof.

That's probably everything read-only, then.  Maybe include an "expert" mode that can be run to unlock the disk but with big warnings: "Proceeding risks file system corruption!"

RNCTX:
Hey guys, done a lot of playing with this for my own images and put up a fork with the steps I used to make mine, and happened to deal with alot of the same issues when messing with FreeNAS over the last few years, which also uses unionFS to run the OS from a read-only flash drive.

Specifically...the following guide has a good idea..

https://hallard.me/raspberry-pi-read-only/

Aliasing commands in the system-wide shell profiles (and the root profiles for things run as root) will leave them executable system wide.  In the read only images I made/forked I did that per that guide.

Giving a modification or custom script write permission is then as simple as adding a "rw" at the top and a "ro" at the bottom of the script, and changing config paths in scripts to point to the permanently stored etc rather than /etc.

From that point, the only hurdles seem to be apply_noobs_os_config and regenerate_ssh_host_keys, the run-once scripts in init.d.  The first applies locale and other such settings chosen in raspi-config, and the second generates ssh keys on the first boot and starts sshd.  They both delete themselves after running once.  When I executed those manually in a chroot on another machine after building the images, the images booted normally with everything running normally.

kdknigga:
I like the idea of being able to toggle between ro and rw.  The question, I guess, is, "is this idiot proof enough?", as it still does make it possible to have interrupted writes and therefore corruption.

bradanlane:
I'll have to load an old RPi project I created 18 months ago. It did everything we are describing here - booted RO, had a configuration script that started with the RW commands, walked the user through setup and customization, then restored RO. There are quite a few of that system deployed to non-technical users. For the developer or expert, the RW and RO scripts were available to toggle the system at anytime without the need to reboot.

I'll try to get to one of those images tomorrow.

haykinson:
I think there will be two types of users: people who will plug things in and expect them to work, and developers who understand the risks and don't mind corrupting the card too much. I would recommend that if we choose to make the default read-only then we just add instructions or a single command to remount everything as read-write. Developers who care will create the image, log in, flip things to read-write and be done with it. Everyone else will just use the web UI.

RNCTX:
I agree, anyone who wants to SSH in and mess with stuff should be perfectly capable of working in /etc_perm instead of /etc, and toggling read/write as they go.

For 99% of users they'll never see the boot screen.

So with that in mind it seems from my tinkering that all you need to do is rewrite apply_noobs_os_config and regenerate_ssh_host_keys, have your dev guidelines point all config changes to the permanent etc location with read+write and read-only toggles at the top and bottom of any scripts, then you're done.

haykinson:
@RNCTX I guess I meant something different -- I think that anyone who is able to SSH should just have a way to flip the card to read-write permanently so that there's no need to much around with odd directories.

RNCTX:
@haykinson I was editing as you were typing I think.

From my experience with unionfs and its alternatives, it solves this problem and makes the changes invisible to the OS, but not so to the admin or the developer.  By nature it involves moving things, particularly /etc.

As I mentioned above a master 'make-read-only' script for generating release images might be a better idea.  That way people working on the code don't have to fight the directory nuances at all, and if users want to do so they can figure it out on their own.

bradanlane:
I'll definitely have to review the current Stratux approach for RO vs what I have from my past project. Either my past project missed something, or there is a way to make things RO without moving things to special directories.

Again, I won't get to this until tomorrow.

kdknigga:
The only thing that I can think would really require the unionfs bit would
be /var.  Everything else *should* be able to live read-only for day-to-day
use (assuming /tmp on tmpfs).

RNCTX:
/etc is included for mtab and resolv primarily.  resolv is the live DNS info and mtab is the live mount status.

kdknigga:
resolv.conf isn't a problem.  The rpi generally won't have internet access,
so why does it need DNS servers configured?  Also, the only time
resolv.conf is dynamic is on a dhcp client.  In practice, it should be fine
to statically set 8.8.8.8 and 8.8.4.4 as the name servers.

mtab is interesting.  I forgot about that guy.  The internet, though,
suggests you can symlink to /proc/self/mounts.

RNCTX:
I verified that you can in fact do that yesterday.  Built a read-only image from the latest release with qemu on another machine, and as the last step just deleted mtab and linked it to proc's mounts, and it booted fine.

jpoirier:
Having the option to build a writable image would be useful.

RNCTX:
To clarify for those unfamiliar with the intricacies of unionfs...

Modified parts of the file system are broken into three parts, think of them as a symlink going in three directions.  For instance /etc becomes /etc_perm, /etc_readwrite, and a new /etc.  The perm one is saved on the system disk and persistent.  The readwrite one is intended to be in a location other than the read-only main system disk typically (not really applicable to this project) and writeable.  And the default /etc that daemons and such look for is a seamless merge of the other two, mounted in RAM during boot.

So in the NAS example mentioned above, /etc_perm is on the bootable read-only system disk mounted read-only, /etc_readwrite is on the user's drive array so that any changes the user makes are saved there permanently.  The primary /etc is a combination of the two generated at boot by unionfs (hence the term union).  For daemons, applications, whatever that write to /etc?  The change is transparent.  Their changes will get written as if /etc is still there in its old form.  But those would need a means of committing those changes to the /etc_perm files if they are intended to be permanent.  Could be a helper script called at the end of each script that needs to write to /etc permanently, or a cron script that diffs and merges changes every so often, or a shutdown script that does the same during the shutdown scripts, whatever. 

In that NAS example unionfs is a perfect drop-in solution to this problem with no 'babysitting' of files or paths required.  A NAS has writable disks all over the place, they just aren't the system disk.  This project is unique, however, and thus will require some 'babysitting' of files and/or paths regardless of which method is chosen to accomplish a read-only system disk, because there is no writeable disk to use.

bradanlane:
is the current thinking that a RO image can not be "toggled" RW and that the only way to make the system RW for short periods (as in to update one arbitrary file) is to have a completely different build?

RNCTX:
The benefit of spindle is it producing 'unbooted' images.  The pi distro has a few files that are modified on the first boot.  The prompt you get on the first boot to expand the SD card, choose your locale, etc is one.  And there's another that generates SSH keys and starts the daemon on the first boot.  Those files delete themselves after they have been successfully run one time, they're in init.d with the other startup scripts.

An easier way to accomplish this stuff would be to abandon spindle and just keep a base development image at all times.  When an image is ready for release, put the rewritten run-once scripts in their places and off you go.

PR #29, opened 2015-09-02
webUI update for time and temp

bradanlane:
changed the UTime to be a value and now handle all formatting in the webUI
render the CPU temp in the webUI


bradanlane:
I updated the version constant in gdl90 but I now see I use the wrong test. "0.2pre+" should be "0.2b1+".

I likely won't matter since you will change it for the very next posted build.

PR #32, opened 2015-09-05
show which types of uat messages we've gotten

haykinson:
known issue with this still: not resetting to zero after a product is no longer sent by the server.

### DIFF
--- a/main/gen_gdl90.go
+++ b/main/gen_gdl90.go
@@ -390,20 +390,28 @@ func parseInput(buf string) ([]byte, uint16) {
 }
 
 var product_name_map = map[int]string{
-	0: "METAR", 20: "METAR", //"METAR and SPECI";
-	1: "TAG", 21: "TAF", //"TAF and Amended TAF";
-	2: "SIGMET", 22: "SIGMET", //"SIGMET";
-	3: "Conv SIGMET", 23: "Conv SIGMET", //"Convective SIGMET";
-	4: "AIRMET", 24: "AIRMET", //"AIRMET";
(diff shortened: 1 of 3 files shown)


bradanlane:
Interesting addition. One thing this code will introduce is a mis-alignment with the responsive layout is side-by-side.

As new content appears in the webUI, that layout is becoming stressed.

The webUI needs to grow beyond its one "status" screen.

The UAT details look like another good example of "details" which need to have a secondary screen. Another example of "details" in development is Log Messages. All "details" should have a consist UX.

I don't know what the right UX is at the moment which is why the Log Messages are not exposed yet. I'll keep the UAT strings in mind when I do the design.

haykinson:
@bradanlane perhaps just inner tabs? this way each aspect gets a tab sheet of its own to play in, and can call the server only when on that sheet?

cyoung:
Good addition, thank you!


"known issue with this still: not resetting to zero after a product is no longer sent by the server."
Can you elaborate on this?

cyoung:
Check extract_metar.go:decodeUplink(), getting an accurate count on this requires that we start parsing the uplink packet in that way. One uplink message will, for example, contain multiple METARs.

haykinson:
@cyoung re not resetting to zero, the UI doesn't clear rows without data, which means it forever sticks around at the last value received. I'll fix that in a separate PR.

Re decoding uplink messages, yes, I think we could get much more sophisticated. I'm not sure that's as important as getting some rough insight into what's being received but I'll look at it at some point.

PR #36, opened 2015-09-11
Adding support for Adafruit Ultimate GPS breakout board

ssokol:
This is a rather hackish branch to replace the Ublox GPS support from the RY835AI with the Ultimate GPS breakout board from Adafruit industries. There are only a few significant changes here:

- Copied ry835ai.go to ultimategps.go
- Replaced the serial port value since the Adafruit board is connected to the raw UART on the Pi
- Changed the NMEA message name since the Adafruit board uses basic NMEA
- Changed the precision in the parser since the Adafruit board is slightly less precise

Note that the ultimategps.go file has not been added to the Makefile and WILL NOT BE BUILT unless you edit your Makefile, replacing ry835ai.go with ultimategps.go.

Also, be aware that I built this against an older version of master and may require some editing to work with the current version.

Long term this really should become an option set rather than a separate build. The admin interface should allow the user to select the serial port, the baud rate and the message set (Ublox vs stock NMEA).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,11 +6,6 @@ all:
 	GOOS=$(GOOS) GOARCH=$(GOARCH) GOARM=$(GOARM) go get -t -d -v ./...
 	GOOS=$(GOOS) GOARCH=$(GOARCH) GOARM=$(GOARM) go build main/gen_gdl90.go main/traffic.go main/ry835ai.go main/network.go main/managementinterface.go main/sdr.go
 
-
-ultimate:
-        GOOS=$(GOOS) GOARCH=$(GOARCH) GOARM=$(GOARM) go get -t -d -v ./...
-        GOOS=$(GOOS) GOARCH=$(GOARCH) GOARM=$(GOARM) go build main/gen_gdl90.go main/traffic.go main/ultimategps.go main/network.go main/managementinterface.go main/sdr.go
-
 test:
 	sh -c true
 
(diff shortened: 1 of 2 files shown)


kdknigga:
This brings up an interesting issue.

I've been working on getting my Dual XGPS160 working with stratux and have
been playing with the idea of using gpsd to manage GPS devices and to be an
abstraction layer of sorts.

Consuming processes can then get normalized position and time data either
via json, shm, or dbus from gpsd.

Does this sound like something worthwhile?

cyoung:
What issues are you running into with the XGPS160? You want stratux to integrate directly with it for the ownship reports and such?

kdknigga:
Yeah, trying to get stratux to integrate with it.

I ran into a similar issue as Steven did where the XGPS160 speaks a
slightly different dialect of NMEA than the Reyax module does.

Also, using gpsd would allow ntpd and stratux to use the GPS simultaneously.
On Sep 11, 2015 09:21, "cyoung" <notifications@github.com> wrote:

What issues are you running into with the XGPS160? You want stratux to
integrate directly with it for the ownship reports and such?

—
Reply to this email directly or view it on GitHub
<https://github.com/cyoung/stratux/pull/36#issuecomment-139559370>.

ssokol:
That could work. I did something similar in the EFIS project I was working on, but I used Redis as the pub/sub engine. There were separate single-threaded data acquisition processes for each data source (ADS-B, GPS, IMU / AHRS, COM radio, air data computer, etc.) all of which were configured as Upstart services, so they would automatically restart in the event of a crash. It made it very easy to integrate new (and redundant) components into the system.

FWIW - it might be better to try to architect the complete Stratux solution using similarly decomposed / distributed elements, all manage by the core gen_gld90 process.

haykinson:
:+1: for the suggestion to decompose some of the data feeds from the core. In theory we can also just use interfaces in golang to accomplish some amount of abstraction / standardization, but it does mean that currently all data feeds need to be written in golang. Using some message passing system can allow people to hack data providers in other languages too.

PR #65, opened 2015-09-28
New REST services with getter/setter for settings

bradanlane:
the new REST services for getting and setting the current set of user accessible settings include:

	UAT_Enabled    bool
	ES_Enabled     bool
	GPS_Enabled    bool
	AHRS_Enabled   bool
	DEBUG          bool
	ReplayLog      bool
	PPM            int

All of the existing REST services have been made to support cross-domain AJAx calling.

The above is in support of the new WebUI (to be submitted as follow-on branch). Once the new WebUI is committed, some of the existing data on the websocket should be deprecated.

### DIFF
--- a/main/managementinterface.go
+++ b/main/managementinterface.go
@@ -7,13 +7,15 @@ import (
 	"io"
 	"log"
 	"net/http"
+	"net/http/httputil"    // used for debugging
+    "reflect"              // used for debugging
 	"strconv"
 	"time"
 )
 
 type SettingMessage struct {
-	Setting string `json:"setting"`
-	Value   bool   `json:"state"`
+	Setting  string `json:"setting"`
+	Value    bool   `json:"state"`
 }
 
 type InfoMessage struct {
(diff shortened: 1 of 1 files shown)


ssokol:
What websocket data are you thinking? While I'm all in favor of REST for some features, I think the websocket is by far the better way to provide regular updates of dynamic data. Things like the enable / disable values are perfect for REST get/set. Sat count, uptime, message count, and even message data are all ideal candidates for the websocket.

bradanlane:
I agree the websocket interface is good for streaming data. 

From the current websocket, I suggest we deprecate the "settings" and keep the "status".

It would be best to use websocket interfaces for weather, traffic, and AHRS/GPS.

Now that I have start/stop listening working for the status, I can repeat the pattern for those other streams.

bradanlane:
I'm going to leave the remainder of the backend work to those more skilled than me. :-)

There are basic REST APIs currently available for traffic and weather. Absent new websocket interfaces, I'll code the WebUI to the existing GET interfaces.

PR #73, opened 2015-10-06
add OwnshipModeS to UI; tweaked Settings field validation

bradanlane:
The OwnShipModeS value is assumed to come in as string (HEX) but goes back out as an integer, this really messes with the normal processing of configuration settings. Since teh HEX string is eas yfor users to enter in the UI, it made sense to preserve that in the UI and in the stratux.conf file. Therefore, it becomes necessary to work wit hthe HEX string in the Go code.

I was able to test all of the GO code with the exception of the makeOwnshipReport();

There was an issue when attempting to clear the WeatherList. It would not allow the list to be reset back to 'blank'. I updated the validation code to handle this. I also now detect field focus changes to trigger validation and storage rather than just triggering from the 'form submit' event.

### DIFF
--- a/main/gen_gdl90.go
+++ b/main/gen_gdl90.go
@@ -177,9 +177,10 @@ func makeOwnshipReport() bool {
 
 	msg[1] = 0x01 // Alert status, address type.
 
-	msg[2] = byte((globalSettings.OwnshipModeS >> 16) & 0xFF) // Mode S address.
-	msg[3] = byte((globalSettings.OwnshipModeS >> 8) & 0xFF)  // Mode S address.
-	msg[4] = byte(globalSettings.OwnshipModeS & 0xFF)         // Mode S address.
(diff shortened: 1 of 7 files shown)


bradanlane:
I was not sure exactly how that code worked. The problem from the end user point of view is they can find a HEX or an OCT representation of their ModeS code. Thus, the UI needs to deal with that. Consequently, the UI interfaces with the stratux.conf file so it too needs to be either HEX or OCT.

cyoung:
Are you validating the hex/oct format and/or length in the iface?

bradanlane:
not currently. the HEX code is validated in handleSettingsSetRequest()

bradanlane:
I'll wait to work on any more UI until you can resolve this pull request.

cyoung:
On second thought, it all looks pretty good - was just a little bit unfamiliar with the way you were doing it. No changes needed.

PR #81, opened 2015-10-14
new gps / ahrs page

bradanlane:
uses a combination of data details and graphical representation
the gps location is depicted on a world map; this could be optimized if Stratux is North America only
The ahrs orientation is depeicted with a webgl generate paper airplane; 
most of the webgl runtime code hs been minified (but not all)

there are a few new files added to the 'www' target in the Makefile

a quick demo can be seen here: https://vimeo.com/142514321


MSGTFE:
I think your link to Vimeo changed. This one works better
https://vimeo.com/142514321

bradanlane:
Corrected link: https://vimeo.com/142514321

PR #101, opened 2015-11-08
Jdp dev

jpoirier:
This is ready to merge. Just reply with LGTM if you want me to do the merge.

### DIFF
--- a/main/gen_gdl90.go
+++ b/main/gen_gdl90.go
@@ -455,7 +455,7 @@ func cpuTempMonitor() {
 					globalStatus.CPUTemp = float32(tInt) // case where Temp is returned as simple integer
 				}
 			}
-		}		
+		}
 
 	}
 }
@@ -708,7 +708,7 @@ type settings struct {
 
 type status struct {
 	Version                  string
-	Devices                  uint
+	Devices                  uint32
 	Connected_Users          uint
 	UAT_messages_last_minute uint
 	uat_products_last_minute map[string]uint32
--- a/main/sdr.go
+++ b/main/sdr.go
@@ -181,16 +181,17 @@ func (u *UAT) sdrConfig() (err error) {
 	//---------- Get/Set Freq Correction ----------
 	freqCorr := u.dev.GetFreqCorrection()
 	log.Printf("\tGetFreqCorrection: %d\n", freqCorr)
-	err = u.dev.SetFreqCorrection(globalSettings.PPM)
-	// librtlsdr returned error code inconsistencies, i.e.
-	// error code -2 means the current value is the same as
+	// avoid librtlsdr error code inconsistency, i.e.
+	// error code -2 means the stored value is the same as
 	// the PPM value we sent (dev->corr == PPM)
-	if err != nil && err != -2 {
-		u.dev.Close()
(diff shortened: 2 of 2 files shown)


bradanlane:
I'm noticing a few things with the new code.

1. Am I correct is seeing dump1090 is now launched from within gen_gdl90? If so, we need to edit the /usr/sbin/stratux script to not *also* launch a copy of dump1090.
2. If dump1090 generates any error messages that do not abort the process, they are lost. This is not a problem under normal operations but users who may experience power related oddities (and we have seen them) will not get any evidence in the log files. The only way to detect the errors are by killing off the stratux process and then manually starting dump1090 manually. This gets back to the prior point that dump1090 now appears to be started inside of the main code.

If I have misread the code, then just comment that these questions are based on bad assumptions.

cyoung:
You're right, addressed first issue in 0cc762fdc2b5eeaada87f298c49df5271f00cc35 just now.

On the second issue, we could redirect stdout and stderr and write to another stratux-____.log, that way should be even better considering it's only writing to terminal currently.

kdknigga:
#1 appears to be correct.  I found the same and removed the dump1090 call
from the start-up script on my rig.

I logged 4.7 hours yesterday running the code from this branch and
everything seemed OK.

bradanlane:
I guess I was a littel slow on #1 :-)

I jsut added a #3 case - "if the UAT SDR is disconnected"

cyoung:
Nope, I was just fast :)

bradanlane:
found the "sdr unplug" problem. In (UAT) read()

```
	for {
		select {
		default:
			nRead, err := u.dev.ReadSync(buffer, rtl.DefaultBufLength)
			if err != nil {
				log.Printf("\tReadSync Failed - error: %s\n", err)
				break
			}
			// log.Printf("\tReadSync %d\n", nRead)
			if nRead > 0 {
				buf := buffer[:nRead]
				godump978.InChan <- buf
			}
		case <-uat_shutdown:
			log.Println("UAT read(): shutdown msg received...")
			return
		}
	}
```

The "break" will only break from the `select` and not from the outer loop.

cyoung:
Make sure you

     go get -u all


or you'll get errors like:

     2015/11/10 23:47:13 	SetSampleRate - rate: 2083334
     2015/11/10 23:47:13 	GetSampleRate: 2083334
     2015/11/10 23:47:13 	GetXtalFreq - Rtl: 28800000, Tuner: 28800000
     2015/11/10 23:47:13 	SetXtalFreq - Center freq: 28800000, Tuner freq: 28800000
     2015/11/10 23:47:13 	SetCenterFreq 978MHz Successful
     2015/11/10 23:47:13 	GetCenterFreq: 978000000
     2015/11/10 23:47:13 	Setting Bandwidth: 1000000
     2015/11/10 23:47:13 	SetTunerBw 1000000 Successful
     2015/11/10 23:47:13 	ResetBuffer Successful
     2015/11/10 23:47:13 	GetFreqCorrection: 0
     2015/11/10 23:47:13 	SetFreqCorrection 0 Failed, error: invalid parameter(s)
     2015/11/10 23:47:13 UATDev = &UAT{indexID: id} failed: invalid parameter(s)
     2015/11/10 23:47:14 DeviceCount... 1

PR #104, opened 2015-11-11
s/return/break and comment out noisey log message

jpoirier:


jpoirier:
It needs to shutdown gracefully via the following method

func (u *UAT) shutdown() {
	log.Println("Entered UAT shutdown() ...")
	close(uat_shutdown) // signal to shutdown
	log.Println("UAT shutdown(): closing device ...")
	u.dev.Close() // preempt the blocking ReadSync call
	log.Println("UAT shutdown(): calling uat_wg.Wait() ...")
	uat_wg.Wait() // Wait for the goroutine to shutdown
	log.Println("UAT shutdown(): uat_wg.Wait() returned...")
}

bradanlane:
When I pulled the UAT SDR I saw "ES" shutdown messages. When I plugged the UAT SDR back in I saw dump1090 try to start (but it was already running so I ended up with defunct processes).

jpoirier:
As far as the zombies go, it probably has something to do with Go's cmd.Process.Kill(). I'll look in to it.

How is the dongle being enabled in the web ui and does it have an id string?

bradanlane:
I don't think I'm explaining this well. 

I unplug UAT SDR and then plug UAT SDR back in. 

The Log is showing it taking actions on dump1090. 

I never touch the 1090ES SDR.

bradanlane:
The WebUI knows nothing about the SDR hardware.  

It only gets/sets settings in the stratux.conf via the REST API.

The WebUI status page is reporting what it gets from the websocket.

skypuppy:
There does appear to be some cross-functionality between the 1090 and
978 going on.

jpoirier:
The web ui helps as a sanity check against what the logs are saying since it reports the number of inserted dongles and what mode they're running in. If neither 978 nor 1090 mode is enabled on the web ui, inserting and extracting a dongle does nothing from the sdrWatcher point of view (ie they're not enabled/configured; that happens dynamically when a user enables the dongle), other than write how many dongles there are to the log. 

Note that even though a dongle is inserted the only way it gets configured and used is if either 978 or 1090 mode is enabled-via the web ui. If a dongle is inserted and only 978 is enabled on the web ui but the dongle has a "stratux:1090" ID it won't be enabled. 

So I'm asking what mode was enabled on the web ui when you were seeing dump1090 being spawned and is it possible that the dongle has an ID of stratux:1090 written to the eeprom.

bradanlane:
Both UAT and 1090 were enabled via the "settings" page. When the UAT SDR was unplugged, the status page went from showing 2 SDRs to 1 SDR.

jpoirier:
I realize the config file is used to relay state between the stratux code and the web ui but the transport mechanism is not really important. 

Also, what makes the particular dongle you're using a UAT dongle?

jpoirier:
If you only had a single dongle inserted and the settings page showed two then something is amiss. The sdrWatcher actively queries the hardware for the number dongles every loop and reports the number back..

jpoirier:
If both 978 and 1090 are enabled on the web ui settings and there is only a single dongle inserted then

if the dongle's ID != "stratux:1090": configure as 978
else: configure as 1090

bradanlane:
This appears to be going around in circles. 

I had Stratux settings for 2 SDR. 
I had 2 SDR installed.
I started Stratux.
Both SDR were functioning. 
I unplug one SDR.
It was the SDR used for UAT based on all previous usage in the same USB port.
I plugged the SDR back in.

ReadSync errors occurred as soon as I pulled the SDR.
The WebUI went from 2 to 1 SDR.
The error messages persisted even after I plugged the SDR back in.

jpoirier:
The USB port is irrelevant and the code doesn't work the way it used to.

With two dongles inserted and 978 and 1090 enabled in the settings and neither had an ID string then:

if you pull the dongle that was running as 1090 then dump1090 gets killed and the 978 dongle continues to run.

If you pulled the dongle that was running as 978 then the 1090 dongle gets reconfigured as 978 and to do that the dump1090 process needs to be killed.

Keep in mind that the algorithm needs to make some sort of intelligent decision on how to proceed given dongles with no known ID strings, multiple modes enabled in the settings, and a single dongle. Current it's to default to 978 for that particular scenario.

bradanlane:
OK. Then it appears the logs were representative of what you explain.

The defunct processes need to be cleaned up.

I personally would disagree that the SDRs should reconfigure their purpose if the UAT is removed. Others may or may not agree.

If everyone agrees with the new behavior, let's get it documented so others don't waste time trying toxicity out what going on.

jpoirier:
Yes, the defunct process need to go away.

How would you handle the situation when both 978 and 1090 are enabled in the setting and a user plugs in a single dongle that has no known ID string?

jpoirier:
I wouldn't expect it to be the definitive way of doing it and the algorithm can be tweaked, the tweaking part is to be but expected since it's a first cut.

kdknigga:
> Also, what makes the particular dongle you're using a UAT dongle?

The antenna plugged into it?  I, for one, am using tuned antennas.

bradanlane:
I use tuned antennas as well.

jpoirier:
How did you ensure the freq specific antenna dongle was being configured
properly previously?

kdknigga:
I plugged the dongles in one at a time and set the serial numbers to
stratux:980 and stratux:1090, respectively.

With both plugged in I just got the device ids and set dump1090 to use the
device id that corresponds to stratux:1090.

Maybe I was getting lucky, but it always worked for me.

cyoung:
>Yes, the defunct processes need to go away.

Fixed in 3745a805b7a5ecbbccbf62685bd5fa8c690556f4.

jpoirier:
You shouldn't have any problems with the current code.

If a dongle has a known ID string ("stratux:1090" or "stratux:978") it is
always configured based on the string, provided that mode is enabled:
"stratux:1090" -> 1090ES mode, "stratux:978" -> 978UAT mode.

It's ambiguous though when a dongle doesn't have a known ID string and both
modes (978UAT and 1090ES) are enabled in the settings.

jpoirier:
The thought was that it's fairly easy to disable/enable modes, so for the
ambiguous case when a user has both modes (978UAT and 1090ES) enabled in
the settings and two dongles (neither with known ID strings) inserted and
then pulls one dongle out.

So, with both (978UAT and 1090ES) modes enabled:

1) If the user pulls out the 1090 configured dongle the remaining (978)
dongle stays configured in 978 mode. If the user wants the remaining dongle
to be 1090 he/she can disable 978 mode and the dongle will re-configure to
1090.

2) If the user pulls out the 978 configured dongle the remaining (1090)
dongle gets reconfigured to 978 mode. If the user wants the remaining
dongle to be 1090 he/she can disable 978 mode and the dongle will
re-configure to 1090.

When there are no ID strings available, it really comes down to proper use
of the mode settings and the user being explicit about what mode to use for
the given resource.

skypuppy:
I agree with the premise that 978 should always have priority for us
'ground huggers.'  (below 18,000 feet)

jpoirier:
It basically comes down to how to handle a single dongle that has no know
ID string when both 978 and 1090 modes are enabled in the settings. The
current code doesn't care how you got/get to that state but it's outcome is
always deterministic, ie the same, config for 978. Adding code that's
cognizant of the system's state prior to the transition to a mufti-mode
setting with a single dongle state means more complexity in the code, but
it's certainly doable.

The final state of all three of these scenarios ends up being the same,
both modes enabled with single dongle, no known ID string:
- both 978 and 1090 modes enabled, no dongle inserted, user inserts a
single dongle (no known ID)
- both 978 and 1090 modes enabled, one dongle inserted (no known ID), user
powers up the system
- both 978 and 1090 modes enabled, two dongles inserted (neither with known
IDs), user pulls out one dongle

We can add a specific/different outcome for each of the listed cases but my
worry is that the code will be more convoluted and/or complex than it
already is for marginal gain considering how easy it is for the user to
just tap a button on the web ui settings to disable one of the two modes,
in which case it makes this a mute discussion.

bradanlane:
I also agree:

If a user has both 978 and 1090 turned on
And only one SDR is present
And there is no distinguishing SDRs  
Then Stratux should **startup** in 978.

I don't feel it's a good idea to "switch on the fly".

Again, this is just one opinion.

jpoirier:
Unfortunately, there are several moving parts required by the housekeeping
code that allow a user to hot insert and extract dongles at will and they
can easily get out of synch, with volatile results. And as stated
previously, the complexity of the code would increase significantly to
handle the one corner case; and to be clear it is a corner case.

With that said, I'm not opposed to changing the-currently very well
defined-behavior, but I am disinclined to spend an inordinate amount of
time on a corner case feature that can easily/simply be solved by the user
properly/explicitly selecting the desired mode via the settings.

Furthermore, I'm guessing that when a user has two anonymous (ie no known
ID string) dongles inserted they don't know which dongle is running in 978
mode and which one is running in 1090 mode. Unless of course the user
inserts and enables each dongle individually one at a time, in which case
the preferred method of designating the mode of a specific dongle should be
via ID string.

Will a reasonable person treat the dongles as they would a USB storage key?
Probably not, but we do need to handle the situation gracefully, and with
well defined behavior, when they do, regardless of what that behavior is.

bradanlane:
While the current behavior my is now well defined, it is very different from what Stratux has done for the past 5+ releases. With more than 800 downloads, we have no idea how many real users we have.

We can't release the new behavior without clear user documentation for setting up Stratux for this new behavior and explaining the behavior if they temporarily an SDR,

My concern is the new behavior is not intuitive, given how all of the currently released versions of Stratux work.

skypuppy:
Whether an owner removes a dongle or a dongle fails in operation, I am
at a loss as to why this is such a difficult situation.

David

jpoirier:
This is alpha/beta software, there have been no major releases nor have
there been any documented guarantees concerning the functionality. So the
disruption argument isn't an orthogonal one and I'd almost go as far as
saying it's risible considering the age and state of the project, but I
won't.

Please point me to the current normative, and not informative,
documentation and I'll be happy to add the current behavior.

As stated previously, it's a corner case that I doubt will occur that
often, etc...ad nauseam

bradanlane:
OK, so we have over 800 beta users.

There is lots of "documentation" on reddit including how users can identify which SDR is servicing 1090 vs 978.

jpoirier:
Yes, I agree that 800+ beta testers is a lot but it is beta software and
making good design/architecture decisions shouldn't be based on the tester
count, at least not at this stage/age of the project. There was no real
design to the previous code. it lacked functionality and had some serious
flaws, no disrespect to Chis, so it's an unfortunate situation that
potentially many users have become comfortable with how it worked.

My point is that any current documentation, where ever it lives, is
informative and not normative.

jpoirier:
On Wed, Nov 11, 2015 at 11:40 AM, Joseph Poirier <jdpoirier@gmail.com> wrote

>
> The current code is simple and does a good job of keeping the housekeeping
> in order (so far anyway) for anonymous dongles but there's an edge case
> that when both modes
>
> Sorry, "edge" case should be "corner" case.

PR #107, opened 2015-11-11
avoid setting frequency correct to zero to avoid eronious error state

bradanlane:
fix bug where SetFreqCorrection (0) generated error which propigated back to sdrConfig to fail which caused the whole process to repeat.

Also prefixed UAT related messages with "UAT". This will help with log interpretation


jpoirier:
This has already been fixed in the gortlsdr library. Scan the previous
commits.

jpoirier:
sdrCount doesn't exist on tip.

PR #112, opened 2015-11-12
godump978 use system lib path, copy dump978 so before it's needed, fi…

jpoirier:
…xes for circleci builder

The ci builder appears to work now.


jpoirier:
yup. it works.

Issue #68, opened 2015-09-30
Request: project status / milestones in Readme


egid:
I (and others, mostly on Reddit) have occasionally gotten a bit confused about the state of various features, especially AHRS. I've got a pretty good read on project status now, but I know others still aren't clear on things.

Might help avoid confusion if there was a list of things that work, mostly work, are in progress, on the roadmap but unstarted, that sort of thing.

bradanlane:
@all - do we want to plan a release with what has been accomplished since 4r4?

Are there fixes or features we should get wrapped up?

_(I tend to keep tweaking stuff until there is a deadline)_

egid:
I can't speak to that, but this is addressed by #123!

@bradanlane, any interest in another js dev working on the web ui? I'd like to help out, and that's my area of expertise.

bradanlane:
@egid - stratux is "open source" so anyone is welcome to contribute. My suggestion is to do you dev in a branch or your own fork; then make a pull request.

I was about to work on a little refactoring of the code to bubble up the common settings so the various pages can visually "do the right thing". 
* weather page should display informational message when stratux configured for 1090 only
* traffic page should display informational message when stratux configured for 1090 only
* gps/ahrs page should display just gps, or just ahrs, or just informational message with stratux is configured for just gps, just ahrs, or neither respectively

I have not dont this yet but I did create bug #120 to track it.

If you are familiar with AngularJS, I would appreciate a code review to best practices since I was "learning on the fly" for this project.

egid:
I'm only passingly familiar with angular - the majority of my webapp work was Dojo, and now React - but I'll take a look.

I mostly want to make sure that if I'm going to mess around with stuff, I'm not stepping on work in progress. Once I get oriented in the code I can grab an issue and see where it goes.

bradanlane:
I railed against Dojo after far too much of it for a company I use to work for :-)

As for React, it was a tossup between React and Angular and I tossed a coin :-)

egid:
Oh yeah, Dojo is awful. React + Flux is quite nice! We needed the ability to quickly render predominately-js pages server-side, so React it was.

cyoung:
>@all - do we want to plan a release with what has been accomplished since 4r4?
>
>Are there fixes or features we should get wrapped up?
>
>(I tend to keep tweaking stuff until there is a deadline)

[...]

PR #123, opened 2015-11-20
Website intro & status sections

egid:
Roughed out the intro and feature / EFB support status sections. Addresses #68. 

This was originally autogenerated by gh-pages but I'd like to come back around for another pass and do it a little less awkwardly.


cyoung:
I want to add some EFBs into the table, do you have any guidelines on changing the page(s) to keep it editable by you?

egid:
So, I'm comfortable editing it raw, and the way the generated pages keep stomping on stuff seems unmaintainable.

I'm going to spend today thinking about better ways & messing around with my local copy, but in the meantime, have at it in any way you like. I'll pull the changes in and when I take another formatting pass I'll make sure they're kept.

Issue #130, opened 2015-11-28
Feature request - Port Number


Axtel4:
Please consider adding an option to configure the port number from the Wed UI or allow more than one port “out of the box” for different Apps.  I know this was broached in Issue 61 and was subsequently closed, but I think it may be worth looking at it again.

In instances where a Stratux device is deployed with multiple users (i.e. within a flying club), there is an incompatibility with some apps at the default port settings. This requires the user to manually change the config file prior to using Stratux. This could lead to someone inadvertently disabling Stratux for a subsequent user.

For example App A works out of the box with Port 4000, however App B requires Port 43211. User 1 has App A and User 2 has App B. User 1 will be able to use Stratux “out of the box” with the default setting. Whereas, User 2 would need to edit the config file for Port 43211 prior to using Stratux for their flight. This forces user 2 to know how to manipulate Linux and edit the files. In addition, there is a risk that if User 2 does not return Stratux to Port 4000 when completed with their flight and when User 3 comes along to use Stratux with their App C they find that Statux no long works.

skypuppy:
Just set software switches, and let them choose which hardware (by name)
that they want to emulate.  They don't have to know anything about
computers to use them.  "Just choose the box on this page," with the
default setup as what everyone else uses.

Easy peasy.

cyoung:
Which app accepts GDL90 on port 43211 but not 4000?

Axtel4:
DroidEFB. I have only gotten it to work with {"Conn":null,"Ip":"","Port":43211,"Capability":5} or with {"Conn":null,"Ip":"","Port":43211,"Capability":1}. When tried with "Port":4000, Droid EFB shows "ADSB NO TRAFFIC TARGETS" on the chart page and no Targets are displayed. Switching back to "Port":43211 DroidEFB shows "ADSB TRAFFIC DETECTED" with Targets being displayed.

cyoung:
Going to call this a DroidEFB bug, then. It's the only app that I know of currently using 43211 and not 4000.

cyoung:
@Axtel4 - still nothing back from DroidEFB. Not going to do this because 1) there is only one app that I know of that is using 4000 only, 2) they probably will support 4000 + 43211 at some point in the near future, 3) you can easily fix it with a config change below (and not ever have to change between the two).

The drawback to fixing with the config is that reduces the number of clients you can connect... so maybe 10 clients instead of 20 :)

[...]

PR #134, opened 2015-12-06
User configurable UDP port in settings menu

jamez70:
I have to test how an old configuration file is handled by this new code

### DIFF
--- a/main/gen_gdl90.go
+++ b/main/gen_gdl90.go
@@ -104,6 +104,7 @@ type ADSBTower struct {
 	Messages_total              uint64
 }
 
+
 var ADSBTowers map[string]ADSBTower // Running list of all towers seen. (lat,lng) -> ADSBTower
 
 // Construct the CRC table. Adapted from FAA ref above.
@@ -378,6 +379,7 @@ func heartBeatSender() {
 func updateMessageStats() {
 	t := make([]msg, 0)
 	m := len(MsgLog)
+	
 	UAT_messages_last_minute := uint(0)
 	ES_messages_last_minute := uint(0)
 	products_last_minute := make(map[string]uint32)
@@ -737,6 +739,7 @@ type settings struct {
 	PPM            int
 	OwnshipModeS   string
 	WatchList      string
+	UDPPort	       uint32 
 }
 
 type status struct {
@@ -769,6 +772,7 @@ func defaultSettings() {
 	globalSettings.DEBUG = false
 	globalSettings.ReplayLog = false //TODO: 'true' for debug builds.
 	globalSettings.OwnshipModeS = "F00000"
+	globalSettings.UDPPort = 4000
 }
 
 func readSettings() {
--- a/main/managementinterface.go
+++ b/main/managementinterface.go
@@ -179,6 +179,9 @@ func handleSettingsSetRequest(w http.ResponseWriter, r *http.Request) {
 							globalSettings.ReplayLog = v
 							replayMark(v)
 						}
+					case "UDPPort":
+						globalSettings.UDPPort = uint32(val.(float64))
+						restartNetworkConnections()
 					case "PPM":
 						globalSettings.PPM = int(val.(float64))
 					case "WatchList":
--- a/main/network.go
+++ b/main/network.go
@@ -151,6 +151,10 @@ func refreshConnectedClients() {
 		return
 	}
 	dhcpLeases = t
+
+	// Use the UDP port specified in the configuration
+	globalSettings.NetworkOutputs[0].Port = globalSettings.UDPPort;
+	
 	// Client connected that wasn't before.
 	for ip, hostname := range dhcpLeases {
 		for _, networkOutput := range globalSettings.NetworkOutputs {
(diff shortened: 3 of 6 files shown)


cyoung:
Deferring, not sure we really want a configurable port.

Axtel4:
This appears to be related to the discussion in Issue https://github.com/cyoung/stratux/issues/130 and was eventually designated as an "externalbug".   

Also, this may be what @skypuppy meant when he said: "Just set software switches, and let them choose which hardware (by name) that they want to emulate.  They don't have to know anything about
computers to use them.  "Just choose the box on this page," with the default setup as what everyone else uses."  Having a Hardware / systems background and not having much software experience I was struggling on how to accomplish this.  

I am interested in testing this on my Startux build, but I am not sure how to pull the files over to my R-Pi to make a build based on these changes.  Where we have this deployed for a Flying Club, we have to keep manually switching from Port 4000 for the iPad users to  Port 43211 for Android users before Stratux can be seen by the respective App.  What we have found is that there are some apps that use Port 4000 (Stratus approach?), while there are some apps that use the iLevel / SkyRadar approach and use Port 43211. There are also seem to be some Apps that don't appear to care and will utilize either port.  Doing this through the Web UI will make it easier for all users to switch the Port on the fly.

It all seemed to work quite well until Version 0.4r2 (Issue 61 / Commit 202e36c3a4c7e05499393e04f22b05d03457f509) where FlyQ started having trouble with being presented with 2 Ports containing data and Port 43211 was removed as a default Port.

cyoung:
@Axtel4 - you can set the config file such that it manages both ports. I've eliminated the other port because, as far as I know, there is only one app that uses 43211 and not 4000 (DroidEFB).

cyoung:
Are your Android users using something other than DroidEFB that has a problem with port 4000?

Axtel4:
I could update our config file to utilize both ports, but won't that break FlyQ for those users since it will now be presented with active two Ports from Stratux?

From issue 61:
"We're going to go with 4000 and eliminate 43211. The above config still works as a temporary fix for now, which is using 43211, but it doesn't really matter much for FlyQ - as long as it's just a single port."
. . . and . . .
" We’re just talking about out of the box experience as FlyQ EFB does currently have a problem if the Stratux transmits on both ports and correcting it is delicate work while changing a config file takes 1 minute. "

Thanks.

cyoung:
"maybe", they've made some changes since then. I still question the utility of using both ports at once. Who is going with 43211 only? Other than DroidEFB

Axtel4:
DroidEFB as far as I know.  So, what can I do to make this a seamless transition for our users as they  move from App to App and platform to platform without having someone "on call" to reconfigure Stratux each time?  (Short of having two Stratux in the aircraft.)

rgutlon:
Axtel4:

The DroidEFB FAQ indicates:
>We are currently supporting Baron’s MobileLink product for weather delivered over XM Radio by WxWorx, SkyRadar and iLevil via WiFi for in flight weather through ADSB transmissions and the Dual XGPS170 via Bluetooth.

From some quick searches, I found a number of posts where DroidEFB users indicate they enter "Stratux into the SSID field for ilevil or SkyRadar."

Have you tried this as a workaround to the port issue?

Axtel4:
Thanks. I was the one that posted the DroidEFB information over on Reddit.

https://www.reddit.com/r/stratux/comments/3ptjgr/droidefb_and_avare/cw9mn92

As an aside, here is an interesting comment on the iLevel sight concerning the use of Port 49002 for AHRS data. 
" * The iLevil 2 is NOT compatible with ForeFlight.  The iLevil 2 Series can be configured to use ForeFlight's Simulation protocol, which allows you to drive the App with limited functionality. Synthetic Vision, GPS Navigation and Traffic display are available when using an iLevil2. Weather information will NOT be available when using the iLevil2 in combination with ForeFlight, nor does ForeFlight offer support for the iLevil 2 series. When using the iLevil 2 in "ForeFlight Sim" mode, compatibility with all other Apps (except for AHRS Utility V3.0) is suspended, until the iLevil is configured for "default GDL90" mode."

This might give some insight to the Forflight/AHRS issue.

skypuppy:
I can't think of anything, but is there any way the stratux can sense
what the receiver wants/needs?  If so, it could adjust itself
accordingly.  However, we may be back to the default and human-chosen
"other" system in the web interface.  Or something like that.

Aside: Please, guys, I'm not dictating rules or methods here, just
throwing out ideas and sometimes desires.  I've been a software weenie
since 1972 and while I still have a lot to learn, I have crossed these
kinds of UI events before.  :)

Skypuppy

cyoung:
@Axtel4 - DroidEFB is the last one using 43211 and not 4000. They emailed me and said they've ordered the parts and are testing.

Axtel4:
Cool. I was in the process of drafting an email to them about their AHRS support when I saw your latest update.

Axtel4:
> DroidEFB is the last one using 43211 and not 4000. They emailed me and said they've ordered the parts and are testing.

Have you heard anymore from the folks over at DroidEFB?

cyoung:
@Axtel4 - Jan 23, got an email stating that they have been busy and will update when they start working on it. Nothing further. Going to close this now, it seems they'll deal with it at some point but they are very slow. We've spent far more time talking about a workaround than it is worth.

If it's still an issue for you - send me an email. We can just edit the DroidEFB binary to use a different port and be done with it.

Axtel4:
I'm OK with that. I modified stratux.conf to use Port 43211.

PR #144, opened 2015-12-13
Switched logs back to uncompressed. Replay now supported with command…

jamez70:
… line arguments. specify -h for help with arguments


jamez70:
I added the 'developer' flag, which sets the logs to uncompressed

PR #152, opened 2015-12-19
Shutdown and Reboot functions

braddward:
Added a couple of functions to handle rebooting and shutdown of the PI from the status screen.

This is a first attempt at contributing to a project so I wanted to keep it simple in the hopes that I did this correctly.

### DIFF
--- a/main/managementinterface.go
+++ b/main/managementinterface.go
@@ -10,6 +10,7 @@ import (
 	"log"
 	"net/http"
 	"time"
+	"syscall"
 )
 
 type SettingMessage struct {
@@ -228,6 +229,20 @@ func handleSettingsSetRequest(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func handleShutdownRequest(w http.ResponseWriter, r *http.Request) {
+	const LINUX_REBOOT_CMD_POWER_OFF= 0x4321FEDC
+	syscall.Sync()
+	syscall.Reboot(LINUX_REBOOT_CMD_POWER_OFF)	
+}
+
+
+func handleRebootRequest(w http.ResponseWriter, r *http.Request) {
+	const LINUX_REBOOT_CMD_RESTART = 0x01234567
+      	syscall.Sync()
+	syscall.Reboot(LINUX_REBOOT_CMD_RESTART)
+}
+
+
 func managementInterface() {
 	weatherUpdate = NewUIBroadcaster()
 	trafficUpdate = NewUIBroadcaster()
@@ -264,6 +279,8 @@ func managementInterface() {
 	http.HandleFunc("/getTowers", handleTowersRequest)
 	http.HandleFunc("/getSettings", handleSettingsGetRequest)
 	http.HandleFunc("/setSettings", handleSettingsSetRequest)
+	http.HandleFunc("/shutdown", handleShutdownRequest)
+	http.HandleFunc("/reboot", handleRebootRequest)
 
 	err := http.ListenAndServe(managementAddr, nil)
 
--- a/main/managementinterface.go
+++ b/main/managementinterface.go
@@ -219,7 +219,7 @@ func handleSettingsSetRequest(w http.ResponseWriter, r *http.Request) {
 						log.Printf("handleSettingsSetRequest:json: unrecognized key:%s\n", key)
 					}
 				}
-				saveSettings()
+			saveSettings()
 			}
 		}
 
@@ -230,16 +230,14 @@ func handleSettingsSetRequest(w http.ResponseWriter, r *http.Request) {
 }
 
 func handleShutdownRequest(w http.ResponseWriter, r *http.Request) {
-	const LINUX_REBOOT_CMD_POWER_OFF= 0x4321FEDC
 	syscall.Sync()
-	syscall.Reboot(LINUX_REBOOT_CMD_POWER_OFF)	
+	syscall.Reboot(syscall.LINUX_REBOOT_CMD_POWER_OFF)	
 }
 
 
 func handleRebootRequest(w http.ResponseWriter, r *http.Request) {
-	const LINUX_REBOOT_CMD_RESTART = 0x01234567
       	syscall.Sync()
-	syscall.Reboot(LINUX_REBOOT_CMD_RESTART)
+	syscall.Reboot(syscall.LINUX_REBOOT_CMD_RESTART)
 }
 
 
--- a/web/plates/status.html
+++ b/web/plates/status.html
@@ -85,4 +85,6 @@
 			</div>
 		</div>
 	</div>
+	<a class="btn btn-warning" href="/reboot">REBOOT</a>
+    <a class="btn btn-danger" href="/shutdown">SHUTDOWN</a> 
 </div>


cyoung:
Brad, thanks for sending this in. Looks great. Have you tried out the "shutdown" feature? I never shut down my Pi properly, just pull the plug. Does it actually stay "shut down" when you do so?

braddward:
Thanks for the tip on using the built-in syscall exported constants.  I'm pretty new to go and learning my way around.

I made your suggested changes and updated that in my branch.  Do I need to reissue the pull request?

Chris - yes, I've tested this on my device and the shutdown works (you must disconnect and reconnect power for it to startup again after a shutdown.  Reboot also seems to work as expected.

I always was worried about file system corruption just yanking the power, but it never seemed to cause damage.

cyoung:
Nope, looks like you've got it.

jpoirier:
> Thanks for the tip on using the built-in syscall exported constants.  I'm pretty new to go and learning my way around.

Np and welcome aboard!

AvSquirrel:
*"... the shutdown works (you must disconnect and reconnect power for it to startup again after a shutdown. Reboot also seems to work as expected."*

@braddward - welcome! If you need a more elegant solution for startup, break out the soldering iron and wire a momentary switch to connect the "Run" pin to the conveniently adjacent ground.  Shorting "Run" to GND will boot the Pi (if it's off), or perform a hard reset (if it's running).

rpi-2-run-reset

skypuppy:
Is that a hardware reset?  If so, that would be the same as pulling the
power, leaving the file system corrupted.

What would be *real* nice is a method of gracefully shutting it down on
loss of external power, and a restart upon getting external power.

braddward:
This could be a solution.  You could shutdown either via the web interface or by wiring a momentary switch to one of the gpio pins.  The code could monitor for the gpio  being activated and then calling the shutdown function.

http://piplay.org/forum/discussion/564/shutdown-button-using-gpio-pin/p1

jtremolo:
As far as power-up and shutdown via GPIO, I highly recommend this:
http://www.mosaic-industries.com/embedded-systems/microcontroller-projects/raspberry-pi/on-off-power-controller

It works quite well.  I typically use it in the auto-off configuration.

jpoirier:
I'd posted a gpio solution a while back: "
https://www.reddit.com/r/stratux/comments/3hw2x8/rpi2_power_off_button/"

ssokol:
My opinion only, but...

This is all trying to solve the problem the wrong way. Embedded electronics
(even those built on higher-order platforms like the Pi) don't have a
graceful shutdown procedure. I don't have to gracefully power down the King
KLN-94 GPS, the Garmin transponder, the engine monitor, or the ADS-B Out
transmitter. For that matter, think about non-aviation applications: home
routers, Roku / AppleTV set-top boxes, office phones. Many of these run
Linux. None of them have a shutdown process that's required to avoid
corrupting the boot image.

The real solution here is to have Stratux run in read-only mode using a
ramdisk for ephemeral data - DHCP leases, etc. If you want / need to log,
you can either create a separate partition on the uSD card or use a USB
flash drive.

On Mon, Dec 21, 2015 at 9:51 AM, Joseph Poirier <notifications@github.com>
wrote:

> I'd posted a gpio solution a while back: "
> https://www.reddit.com/r/stratux/comments/3hw2x8/rpi2_power_off_button/"
>
> On Mon, Dec 21, 2015 at 9:38 AM, braddward <notifications@github.com>
> wrote:
>
> > This could be a solution. You could shutdown either via the web interface
> > or by wiring a momentary switch to one of the gpio pins. The code could
> > monitor for the gpio being activated and then calling the shutdown
> function.
> >
> > http://piplay.org/forum/discussion/564/shutdown-button-using-gpio-pin/p1
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <https://github.com/cyoung/stratux/pull/152#issuecomment-166332996>.
>
> >
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/cyoung/stratux/pull/152#issuecomment-166337517>.
>



-- 
Steven Sokol
408 Camelot Drive
Liberty, MO 64068

mobile: +1 816-806-8844
fax: +1 816-817-0441

kdknigga:
Here are my unsolicited, semirandom thoughts on the subject:

This is <1.0 software and a lot of stuff is still changing.

This is a DIY project that encourages tinkering.

I don't think many, if any, people are getting file system corruption from
simply unplugging.

I got the stratux image booting into read-only a while back.  It was a pain
and I wasn't getting any advantage from it, so I stopped fussing with it.

ssokol:
You haven't had filesystem corruption *yet*. It will eventually happen.
Given that Stratux is something that people are using for situational
awareness in flight, I would rather be safe than sorry. Here's a compromise
idea:

Offer a "developer mode" that is fully read / write. Pretty much as things
are now. Also offer a "production mode" which is read only except when
updating the configuration. Default to production mode, but include a
simple UI switch to shift into developer mode.

Perhaps we don't need to do this until we reach the 1.0 mark, but
considering the number of people out there using Stratux in production (the
vast majority of which aren't tinkering) I'm in favor of doing it sooner
rather than later.

On Mon, Dec 21, 2015 at 10:37 AM, Kris Knigga <notifications@github.com>
wrote:

> Here are my unsolicited, semirandom thoughts on the subject:
>
> This is <1.0 software and a lot of stuff is still changing.
>
> This is a DIY project that encourages tinkering.
>
> I don't think many, if any, people are getting file system corruption from
> simply unplugging.
>
> I got the stratux image booting into read-only a while back. It was a pain
> and I wasn't getting any advantage from it, so I stopped fussing with it.
>
>
>
> On Mon, Dec 21, 2015, 10:22 Steven Sokol <notifications@github.com> wrote:
>
> > My opinion only, but...
> >
> > This is all trying to solve the problem the wrong way. Embedded
> electronics
> > (even those built on higher-order platforms like the Pi) don't have a
> > graceful shutdown procedure. I don't have to gracefully power down the
> King
> > KLN-94 GPS, the Garmin transponder, the engine monitor, or the ADS-B Out
> > transmitter. For that matter, think about non-aviation applications: home
> > routers, Roku / AppleTV set-top boxes, office phones. Many of these run
> > Linux. None of them have a shutdown process that's required to avoid
> > corrupting the boot image.
> >
> > The real solution here is to have Stratux run in read-only mode using a
> > ramdisk for ephemeral data - DHCP leases, etc. If you want / need to log,
> > you can either create a separate partition on the uSD card or use a USB
> > flash drive.
> >
> > On Mon, Dec 21, 2015 at 9:51 AM, Joseph Poirier <
> notifications@github.com>
> > wrote:
> >
> > > I'd posted a gpio solution a while back: "
> > >
> https://www.reddit.com/r/stratux/comments/3hw2x8/rpi2_power_off_button/"
> > >
> > > On Mon, Dec 21, 2015 at 9:38 AM, braddward <notifications@github.com>
> > > wrote:
> > >
> > > > This could be a solution. You could shutdown either via the web
> > interface
> > > > or by wiring a momentary switch to one of the gpio pins. The code
> could
> > > > monitor for the gpio being activated and then calling the shutdown
> > > function.
> > > >
> > > >
> > http://piplay.org/forum/discussion/564/shutdown-button-using-gpio-pin/p1
> > > >
> > > > —
> > > > Reply to this email directly or view it on GitHub
> > > > <https://github.com/cyoung/stratux/pull/152#issuecomment-166332996>.
> > >
> > > >
> > >
> > > —
> > > Reply to this email directly or view it on GitHub
> > > <https://github.com/cyoung/stratux/pull/152#issuecomment-166337517>.
> > >
> >
> >
> >
> > --
> > Steven Sokol
> > 408 Camelot Drive
> > Liberty, MO 64068
> >
> > mobile: +1 816-806-8844
> > fax: +1 816-817-0441
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <https://github.com/cyoung/stratux/pull/152#issuecomment-166348455>.
>
> >
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/cyoung/stratux/pull/152#issuecomment-166353244>.
>



-- 
Steven Sokol
408 Camelot Drive
Liberty, MO 64068

mobile: +1 816-806-8844
fax: +1 816-817-0441

bradanlane:
The SDR code is pretty stable. Perhaps the group agrees (or not) that we finish stabilizing the SDR code, add any related "nice to have" stuff, and add read-only support and base line the Stratux 1.0 version.

I'm not suggesting we remove the other functionality; just consider that work "beta" and the SDR work "release".

The follow on work would include any bug fixes for SDR support and focus on the GPS/AHRS features.

Does that sound reasonable to others?

RobertGary1:
My intention is to power my Stratux through the cigar lighter and it will manically lose power when I hit the master. Is that what most envision ? I don't typically do anything elaborate to shutdown my avionics. 

-Roberr

cyoung:
>Perhaps the group agrees (or not) that we finish stabilizing the SDR code

What sort of things do you have in mind re: this?

Read only should definitely be a "1.0 feature", would rather not fork the releases right now because that will slow us down in a number of ways.

bradanlane:
BTW: @kdknigga, I have not looked at your RO implementation but with the exception of DHCP leases, I had an RPi project that was read-only and it was not difficult. It booted as RO and had two command scripts to switchRW and switchRO. The related "switching commands" were used primarily when config file changes were made. eg: switchRW; change conf file; switchRO.

My guess is you had a more robust solution that I did. My solution lost DHCP leases with the reboot since they were mapped to tmp which was in RAM.

bradanlane:
@cyoung - I was not suggesting we fork the code; but instead 

* make a final push on anything needed for SDR support
* add read-only support (and make it what happens when the RPi boots)
* test
* release as 1.0

jpoirier:
On Mon, Dec 21, 2015 at 10:56 AM, cyoung <notifications@github.com> wrote:

> Perhaps the group agrees (or not) that we finish stabilizing the SDR code
>
> What sort of things do you have in mind re: this?
>
> Read only should definitely be a "1.0 feature", would rather not fork the
> releases right now because that will slow us down in a number of ways.
>
I concur that releases, 1.0+,  be read only.

cyoung:
My question would be is if we want to merge everything into a single release image that even beta testers/developers and normal users alike will use, or perhaps cut some features out for the non-developer version. Features like replay logging seem more like a complication on a r/o release for users that likely will never touch it.

bradanlane:
@cyoung - I missed that angle to your question. I would rather not fork the code. Instead, I would use a conditional compile flag to cut the developer realted features out of a real release build.

jpoirier:
On Mon, Dec 21, 2015 at 11:14 AM, cyoung <notifications@github.com> wrote:

> My question would be is if we want to merge everything into a single
> release image that even beta testers/developers and normal users alike will
> use, or perhaps cut some features out for the non-developer version.
> Features like replay logging seem more like a complication on a r/o release
> for users that likely will never touch it.
>
 My gut feeling is that not having diagnostics on a release is a mistake.
Will the removal of features for a release potentially create a large
enough delta between the beta and the release such that the release isn't
representative of what was tested in beta?

jpoirier:
On Mon, Dec 21, 2015 at 11:26 AM, bradanlane <notifications@github.com>
wrote:

> @cyoung <https://github.com/cyoung> - I missed that angle to your
> question. I would rather not fork the code. Instead, I would use a
> conditional compile flag to cut the developer realted features out of a
> real release build.
>
Are you saying to just disable features from the front end, i.e. the web
ui, so the functionality isn't available? Note that Go doesn't have a
pre-processor and it doesn't support any kind of #ifdef.

bradanlane:
I was not aware of that Golang limitation. It would be possible to refactor the code such that the parts were in separate files but that is extra work for little gain. The alternatives are:

* to disable capabilities at the UI
* runtime flag

deandayton:
FWIW, I've created a number of embedded Linux products that by default have a ro root file system. Diagnostics/logging can be enabled via the user interface with sufficient warnings to discourage use by the novice user. Alternatively, a "standard non-developer release" could simply not provide the option to enable diagnostics and logging. It is generally easier to maintain if you leave all of the code for supporting diagnostics/logging in place.

Also note that a separate writable partition for logs has advantages. If it becomes corrupted the system can still boot up normally and even detect/fix the corrupted partition. Worst case you lose the captured logs. I have been involved in shipping hundreds of thousands of systems that operate this way over more than 10 years. These systems have never had the luxury of a guaranteed orderly shutdown.

________________________________

kdknigga:
@bradanlane

Getting RO to work wasn't very hard.  Where I was having problems was using
spindle to generate images.  It took me forever to get it working at all,
and even then it was cumbersome and easy to break.

jpoirier:
On Mon, Dec 21, 2015 at 11:38 AM, bradanlane <notifications@github.com>
wrote:

> I was not aware of that Golang limitation. It would be possible to
> refactor the code such that the parts were in separate files but that is
> extra work for little gain. The alternatives are:
>
>    - to disable capabilities at the UI
>    - runtime flag
>
> Just to expound a bit... Go doesn't support any sort of fine grained
pre-compile switch for lines of code but the Go tool itself does support
file level build tags, e.g. // +build linux, 386 darwin, !cgo. But as you
stated, it would probably only be useful if the code was refactored.

bradanlane:
yes, "spindle" is not the easiest. For my work on the OPi, I created an "update script" that makes all of my necessary changes.

jpoirier:
On Mon, Dec 21, 2015 at 12:03 PM, bradanlane <notifications@github.com>
wrote:

> yes, "spindle" is not the easiest. For my work on the OPi, I created an
> "update script" that makes all of my necessary changes.
>
A script that was able to get stratux setup on live system-run script and
reboot-would be nice.

cyoung:
https://www.reddit.com/r/stratux/comments/3xwpiz/ver_05b2_web_ui_issue/


Any idea?

jpoirier:
The shutdown button was working properly for me, no idea why it's not
working for others. It might be worth trying exec.Command instead:

cmd = "sudo shutdown -h now"
output, _ := exec.Command("/bin/sh", "-c", cmd).Output()
log.Printf("Shutdown cmd output: %s", output)

cyoung:
@braddward - https://vid.me/TKwU

bradanlane:
There are a few tweaks that will make this new feature better. 

1. Instructions for the buttons need to be added to the context help.
2. Confirmation would be nice given the functions being enabled
3. The page redirect needs to happen after conformation and before execution.
4. The interface is more applicable to the "settings" page rather than the status page.


The design of the buttons could use a little UX love too :-)

braddward:
Thanks for the feedback and I agree with all of it.

I may need to add a bootstrap plug in to do the confirmations and feedback.   I'm more of a back-end dev than front-end (hmmm - that doesn't sound so good)...

braddward:
Quick update on the shutdown feature.  I'm refactoring the UI to add modal confirmation dialogs to both shutdown and reboot.  Dealing with learning curve on learning mobile angular JS.   I think I have the hard stuff figured out.
It may be a week or so before I can submit a pull request with the changes.  
Right now I have to make sure I'm ready for my check ride which is hopefully later this next week if I can get it scheduled with the DPE.

RobertGary1:
That's cool of you can get dpe's in a few days. We have to schedule 4-6 weeks out which can be difficult to time with your students. 

-Robert

bradanlane:
As part of the refactoring, did you move the buttons to the "settings" page? I think that's a better place for them. 

I'd like to adjust the styling of the buttons as well but if you're in the middle of changes, I can hold off.

braddward:
Yes, moving to the settings page.  If you want to restyle feel free and I can merge it in.  Otherwise I was just planning on styling them as plain buttons. N

bradanlane:
Thanks @braddward. I'll figure out the design and send you the markup/styling.

braddward:
Well @RobertGary1 - you must have been psychic.  Just heard that the DPE we were going to use next week has an issue and is unable to do check rides....   My CFI is looking for alternatives.

RobertGary1:
Hey, more time to work on Stratux :)

Progress has been amazing. Will be cool to get ahrs working in flight. Not sure if it's a frequency polling limitation or needs a more complex algorithm. I may take a look. It's hard because you can't simulate normal ahrs reaction on the ground. 

-Robert

dbarrington1:
As far as the AHRS goes.  I wonder if Vibration is effecting it.  I found even going into a rather steep climb angle would show a climb and the. Level back off.  Pretty sure I was not in a "coordinated"  climb.  Since there is no such thing in the pitch angle.

skypuppy:
You can on a motorcycle.  :)

PR #162, opened 2015-12-24
Added a NMEA counter to the status to show the number of NMEA message…

jamez70:
…s received


cyoung:
Going to start reducing the amount of data shown on the interface - thanks for sending. Since this we've added GPS fix info.

PR #184, opened 2016-01-03
Updated shutdown

braddward:
Per issue 152 
* moved shutdown and reboot functions to settings panel
* added confirmation modal dialogs to both functions
* added help text


cyoung:
Thanks for the update!

cyoung:
#152

Issue #193, opened 2016-01-08
Request: FE redirect immediately before shutdown


jamesshannon:
Scenario: I want to shutdown. Go to the FE, choose shutdown on my phone, affirm, then it goes to /shutdown and the unit shuts down. I'm lazy, so I don't close the tab.

Later, I bring the stratux back up, open my browser, see that the 192.168.10.1 tab is open and has a "... connection was lost" message. I hit refresh and ... nothing. In fact, the Wifi dongle stops blinking at about this time.

Turns out that the shutdown page was still "open" and I was refreshing that.

 Easy fix: shutdown and restart pages should be POST's, which is at least semantically correct.

Better fix: they should immediately redirect back to the homepage, before the shutdown starts.

Even better fix: homepage (which is already loaded and in the browser) does a XHR POST to the /shutdown page (technically should have a xss token), which causes the shutdown, but never leaves the loaded webapp. The webapp then goes to 'Status: disconnected' and continues to monitor the server until it comes back up.

Nokomis449:
While you're visiting this, please consider moving the reboot and shutdown buttons further apart.  A little turbulence can turn a reboot press into a shutdown press,  which is bad news if the Stratux is on the hat shelf in the back.  I'd like to see them on each side of the div.

ScraboTower:
Please put them back on the first page !!!!  They are technically NOT a setting !!!

bradanlane:
The likely implementation will be AJAX calls to web service a web service "command" interface. This eliminates any change to the URL. The "shutdown" and "reboot" should default the WebUI back to the "status" page.

Adding more space between the buttons is good feedback.

While the "reboot" and "shutdown" buttons are not technically "settings", they are controls and thus are even less about "status".

We debated even having these buttons as they will not be necessary. However, until Stratux is able to provide the stability of an "appliance" (in part by operating with the SD card read-only) there was enough call for the buttons that they were added.

RobertGary1:
The shutdown button is a temporary work around, I wouldn't put too much effort into it. 

Long term  the Stratux image is on a RO file system so you just unplug it. It's an appliance to most users. I've never shutdown mine in fact it powers off when I turn off the master switch.


-Robert

braddward:
Good idea on doing the POST.  I can look at adding that and adding some
spacing between the buttons.

skypuppy:
Robert,  you are not viewing enough of the picture to be  helpful.

bradanlane:
[...]

PR #199, opened 2016-01-09
Updated shutdown

braddward:
Per 193 - modified shutdown and reboot functions to be POST requests.
Put shutdown and reboot buttons in their own areas to reduce chances of accidentally hitting the wrong button.
Added post shutdown/reboot messages.


cyoung:
Fixes #193.

PR #231, opened 2016-02-03
Detect and setup GlobalSat BU-353

iposva:
This is the minimal change to get the puck working on boot. Ideally we would request all of the needed NMEA sentences during the initial configuration instead of relying on the defaults being set properly.

I have not seen a need to request a higher baud rate, so I left that at 4800.


Ergonomicmike:
I've got to take the widow grocery shopping in a moment. So just a quick note for now, and I'll try to develop this later. Is it okay to leave the baud rate at 4800? How would I know in practice?

Typical American thinking is that "bigger is better." ("What do men want?" MORE POWER!!! Grunt, grunt, grunt.) But I presume there is a good reason that the Stratux changes the ublox baud rate to 38400? (Else why go thru the trouble of adding the code?)

Seems like it ought to be a straightforward back of the envelope calculation. So many fields of so many bits (bytes?) sent at say, a worst case of 5Hz, gives a baud rate of?

Having said that, I don't know if the ublox reverts to a slow baud rate after some internal backup capacitor discharges completely. The GlobalSat does. So I suppose that the first handshake should always be at 4800 and then toggle to the faster baud rate next? (Would Chris's Developer Option Hot Start code work here?)

If 38400 is necessary, here's the NMEA command for the GlobalSat:

38400/8/N/1 - $PSRF100,1,38400,8,1,0*3D

iposva:
Maximum NMEA sentence length is 82 characters. We get these messages once a second: GPVTG, GPGGA, GPRMC, GPGSA. This gives us 4*82 = 328 characters, which is under the 480 characters available per second at 4800 baud. To not complicate things I left it at the default baud rate.

If we do switch to a higher frequency of position updates, then we will need to move to a higher baud rate.

Ergonomicmike:
Okay, if I'm doing the math right, 4800 is more than enough. (And I suggest adding a change to increase GGA to 5 Hz.) Here's what I did:

I watched the NMEA sentences for the GlobalSat in SiRFDemo. The longest sentence I saw was 68 characters. Call it 70.

8 bits per character. (?)

70 char *8 biits/char =560 bits

Assume 5 Hz update.

560 bits * 5/sec = 2800 bits/sec. So 4800 baud should allow plenty of head room.

If a character takes 16 bits (there are dollar signs, asterisk, and commas in the strings), then that gives 5600 bits/sec and the baud rate should be moved up to 9600.

Having said that, when the GlobalSat is set to SiRF binary, it ups the buad rate to 57,600. (About 80 characters.) If that's a Sanity Check, then I've done something wrong in my calculations.

Next, I've been told by iFly that 5 Hz update is preferred. Does that mean that GGA should be set to 5 Hz? If yes then,

Enable GGA Output 5sec - $PSRF103,0,0,5,1*25

should be encoded and added.

iposva:
If we switch to 5Hz updates of GGA, then we will need to switch to 9600 baud. What about other data such as VTG?

I'll work on this later today in my fork and test. We can decide whether it needs to be rolled into this pull request or add it in a later stage. The goal of this pull request is enabling the puck in the first place without having to login or change code.

We can create a new pull request for the switch to 5Hz.

Ergonomicmike:
I will email the Sr. Software engineer and ask him what they want for refresh rates. I'll report back.

braddward:
for serial output I assume you're probably setting it to 8 bit, 1 stop bit,
no parity.   So there should be 9 bits per character. Generally this is
also ASCII characters and not unicode so there should only be 8 bits per
character.

jzeevi:
There’s always one start bit. 8-n-1 results in 10 bits per character.

cyoung:
`/dev/ttyUSB0` - is there any other way of detecting this device? I'm actually using the same method to detect a serial out device (see serialout branch), and not sure if there is any other method to detect that one (short of looking at usb strings).

iposva:
@cyoung I have not found a good method yet how to determine which particular device is connected. The USB device tree only lists the USB to serial converter chip as far as I remember. I will take another look at that information.

@jzeevi I generally assume 1 start, 1 stop and 8 data bits. I have not seen 1.5 or 2 stop bits in quite a while and it makes the math easy: Divide by ten.

Ergonomicmike:
On detecting the GlobalSat: Perhaps one can send the "cold start, clear all data" command out on the USB. The response from the puck is "...OK"

From [gpsinformation.org](http://www.gpsinformation.org/dale/nmea.htm)

```
The sentences 101 and 104 can be used to initialize values to be used by the gps. Supplying these values can shorten the initial lock time. If the clock offset is set to 0 then an internal default will be used. Sentence 101 supplies data in the internal ECEF (Earth centered, Earth Fixed) format in meters while sentence 104 supplies the data in the traditional Lat / Lon format.

 $PSRF101,-2686700,-4304200,3851624,95000,497260,921,12,3*22
 $PSRF104,37.3875111,-121.97232,0,95000,237759,922,12,3*3A

where
   $PSRF104
   37.3875111 Latitude in degrees
   -121.97232 Longitude in degrees
   0          Ellipsoid Altitude in meters
   95000      Clock offset
   237759     GPS Time of Week in seconds
   922        GPS Week Number
   12         Channel count (1 to 12)
   3          Reset config where
              1 = warm start, ephemeris valid
              2 = clear ephemeris, warm start (First Fix)
              3 = initialize with data, clear ephemeris
              4 = cold start, clear all data
              8 = cold start, set factory defaults
   *3A        checksum
``

**Comment by Ergonomicmike** [comment]
I'm still waiting on a response from iFly about preferred update rates. In the meantime, since we're disabling the GSV response, which was set to a default of 5 Hz, seems like we can increase GGA by 4 (to 5) and not have changed bandwidth, even with VTG now at 1 Hz. (I've done it on mine.)

**Comment by iposva** [comment]
@Ergonomicmike GSV is once every 5 seconds by default, which is why it did not overly tax the bit rate.

**Comment by skypuppy** [comment]
The faster the baud rate, the greater chance of missing bits, as the Pi
does not use any type of hardware handshaking via the serial port. 
Unknown if the USB port does.

Skypuppy

**Comment by skypuppy** [comment]
One has to be careful of overloading the Pi2, or especially a Pi.  There
is a bit of CPU overhead in handling a serial port that is not being
factored into these posted calculations.  The serial port is polled, and
not using interrupts, making it the most overhead of all for serial
comms.  All that work is handled by the kernel (effectively) but it has
to be done somewhere.

Skypuppy

**Comment by Ergonomicmike** [comment]
I'm confused. In the NMEA settings in SiRFDemo it says that the "update rate is 1/x sec; (0 to 255)." Interestingly, the up/down buttons only scroll to 10.

So I figured that 5 meant 1/5 second. (Although I don't know why one would want to update the Satellite data that fast.) 

You're saying that the integer is a multiplier?

You're correct. I just set GGA to 10 and it only updates once every 10 seconds.

So then we're done? (Except for Chris's request to sniff the GlobalSat different from USB0.) I don't see a way to increase the refresh rate of GGA to 5 Hz. 

If ublox can do it, that's a point in ublox's favor.

Good point @skypuppy that faster isn't always better.

**Comment by Ergonomicmike** [comment]
From the SiRFDemo manual, page 3-43.

"3. Select the update rate for each NMEA message from the corresponding Update Rate
dropdown menu.
Each NMEA message can be output at a maximum rate of one per second and at a
minimum rate of one per 255 seconds."

Rats.

**Comment by iposva** [comment]
@Ergonomicmike You can get at least the GGA at a higher frequency, by using some not very well documented sentence. I managed to set it to 5Hz in the past, but I did not see the immediate value in doing so. I can look it up later tonight and test on my Pi which has the puck connected. I will let you guys know what I find out and post the necessary changes to my repository.

**Comment by Ergonomicmike** [comment]
Okay, I got a response from iFly. Bottom line: He didn't know what was right. Said that 5 Hz sounds right.

(He said that the Stratux is sending GPS at 15 Hz! I wonder if that's the RY835ai board? In any event, he said that 15 Hz is too fast.)

I know that iFly updates their screen once every second. I don't know if they're using any extra data in the background during that time. But I'm thinking that when they get Synthetic Vision going, 5 Hz GPS will be better than 1 Hz. (Might help stabilize the big picture with Kalman Filtering or something.)

So how about this: If you have the time and the inclination, how about putting in a commented Developer's Option for the 5 Hz GGA string. (I suppose that will require a code change to 9600 baud, and then another code change to sniff for a cold start at 4800 default or a hot start at 9600.) If you know the cheat code for 5 Hz on the VTG, you could put that in a commented Developer's Option too.

That way, even if you don't feel like coding the 9600 baud change/check, the not-well-documented commands you found will be preserved for posterity and perhaps, if needed, someone else in the future could use them.

**Comment by skypuppy** [comment]
Or switch to a different gps that publicly supports 1, 5, and even 10 hz.

**Comment by skypuppy** [comment]
Is "ublox"  a brand name and/or a communications protocol?  Thanks.

Skypuppy

**Comment by skypuppy** [comment]
Also, are you guys using the Swiss product ublox-7 or the "low-power"
-8, as seen here:
https://www.u-blox.com/en/press-release/u-blox-releases-u%E2%80%91blox%C2%A08-new-gpsglonass-receiver-platform-low-power-devices

**Comment by Ergonomicmike** [comment]
U-blox is a brand of GPS chipset. See u-blox.com. (I see you found it.) The Stratux officially recommends the VK-172, which uses the u-blox 7 set. It was out of stock on Amazon when I made my order. I opted for the outlier GlobalSat puck intead of a generic no name u-blox based puck. You're correct that, if needed in the future, I could buy a ublox based GPS. (Ironically, the code in Stratux for the u-blox has the baud rate at 38400!)

**Comment by skypuppy** [comment]
Thanks.  Initially, I was under the impression that it was another comms
encoded (not encrypted) message system, like gdl_90.  Reviewing the
stratux code really threw me, especially since I don't know "go."  <sigh>

Without a good IDE/ debugger to step through the code while it's
executing is *really* hampering my efforts at understanding it.

Skypuppy

**Comment by Ergonomicmike** [comment]
Yeah, I'm really resisting trying to read a tutorial on programing in go. I eventually had to teach myself HTML & CSS, but that's because I wanted to create some web sites. I was hoping to use the Stratux w/o getting further than hardware. But I think I see the light at the end of the tunnel. WAAS GPS should be the end for me. Once everything runs w/o disconnecting, I'll call it good. (I'm not interested in AHRS.)

**Comment by skypuppy** [comment]
Ah, but I want it *all!*  :)  :)

**Comment by Ergonomicmike** [comment]
I've never understood the appeal of AHRS. Not for VFR flying anyway. What am I missing?

**Comment by yasbean** [comment]
IFR

**Comment by Ergonomicmike** [comment]
Want something very reliable when I'm shooting my 2nd approach down to minimums. Will take at least a year after final Release before I bet my life on it. (We had the same rule in the Glasair before I flew IFR in it.)

**Comment by AvSquirrel** [comment]
@Ergonomicmike -- situational awareness. When it's after sunset, you're between layers, and hand-flying by reference to vacuum-driven gauges, it's nice to be able to glance down and see the ground.

situational_awareness

@iposva -- there was also a request on Reddit a couple days ago for supporting a third chipset (MTK3339) used in the Adafruit GPS. The monolithic initialization routine is getting a bit top-heavy, and once Chris pulls this into his master branch, I'd like to pull this into my development branch and split up the initialization into separate routines for each chipset.

**Comment by Ergonomicmike** [comment]
Yeah, I get it and it's nice. (Would be especially handy over mountains at night.) I've got an iFly for the right side of your display. As for the left, I've been flying so long that my scan is subconscious. (Not an issue.) My head is so lock in that I've gotten airsick doing rolls by reference to instruments in a non-moving ground sim.

**Comment by skypuppy** [comment]
FUN!

**Comment by iposva** [comment]
@cyoung When dumping the USB devices this is what I get:
root@raspberrypi:/home/pi# usb-devices 

T:  Bus=01 Lev=00 Prnt=00 Port=00 Cnt=00 Dev#=  1 Spd=480 MxCh= 1
D:  Ver= 2.00 Cls=09(hub  ) Sub=00 Prot=01 MxPS=64 #Cfgs=  1
P:  Vendor=1d6b ProdID=0002 Rev=04.01
S:  Manufacturer=Linux 4.1.7-v7+ dwc_otg_hcd
S:  Product=DWC OTG Controller
S:  SerialNumber=3f980000.usb
C:  #Ifs= 1 Cfg#= 1 Atr=e0 MxPwr=0mA
I:  If#= 0 Alt= 0 #EPs= 1 Cls=09(hub  ) Sub=00 Prot=00 Driver=hub

T:  Bus=01 Lev=01 Prnt=01 Port=00 Cnt=01 Dev#=  2 Spd=480 MxCh= 5
D:  Ver= 2.00 Cls=09(hub  ) Sub=00 Prot=02 MxPS=64 #Cfgs=  1
P:  Vendor=0424 ProdID=9514 Rev=02.00
C:  #Ifs= 1 Cfg#= 1 Atr=e0 MxPwr=2mA
I:  If#= 0 Alt= 1 #EPs= 1 Cls=09(hub  ) Sub=00 Prot=02 Driver=hub

T:  Bus=01 Lev=02 Prnt=02 Port=00 Cnt=01 Dev#=  3 Spd=480 MxCh= 0
D:  Ver= 2.00 Cls=ff(vend.) Sub=00 Prot=01 MxPS=64 #Cfgs=  1
P:  Vendor=0424 ProdID=ec00 Rev=02.00
C:  #Ifs= 1 Cfg#= 1 Atr=e0 MxPwr=2mA
I:  If#= 0 Alt= 0 #EPs= 3 Cls=ff(vend.) Sub=00 Prot=ff Driver=smsc95xx

T:  Bus=01 Lev=02 Prnt=02 Port=03 Cnt=02 Dev#=  4 Spd=480 MxCh= 0
D:  Ver= 2.00 Cls=00(>ifc ) Sub=00 Prot=00 MxPS=64 #Cfgs=  1
P:  Vendor=7392 ProdID=7811 Rev=02.00
S:  Manufacturer=Realtek
S:  Product=802.11n WLAN Adapter
S:  SerialNumber=00e04c000001
C:  #Ifs= 1 Cfg#= 1 Atr=a0 MxPwr=500mA
I:  If#= 0 Alt= 0 #EPs= 4 Cls=ff(vend.) Sub=ff Prot=ff Driver=rtl8192cu

T:  Bus=01 Lev=02 Prnt=02 Port=04 Cnt=03 Dev#=  5 Spd=12  MxCh= 0
D:  Ver= 1.10 Cls=00(>ifc ) Sub=00 Prot=00 MxPS=64 #Cfgs=  1
P:  Vendor=067b ProdID=2303 Rev=04.00
S:  Manufacturer=Prolific Technology Inc. 
S:  Product=USB-Serial Controller D
C:  #Ifs= 1 Cfg#= 1 Atr=80 MxPwr=100mA
I:  If#= 0 Alt= 0 #EPs= 3 Cls=ff(vend.) Sub=00 Prot=00 Driver=pl2303

Note how the last device is really just recognized as the USB<->serial bridge. But it gives no indication what is connected to the other side of the bridge chip. To be honest, I am a bit out of ideas how to distinguish between the serial output device and the GPS puck.

**Comment by iposva** [comment]
The latest change adds 5Hz navigation and switches to 38400 baud to be able to comfortably absorb the amount of data. All of the NMEA sentences (GGA, GSA, RMC, VTG) are now sent 5 times a second.

To switch back to 1Hz navigation from the command line:
`echo -en "\$PSRF103,00,7,00,0*22\x0d\x0a" > /dev/ttyUSB0`
To switch back to 4800 baud on the command line:
`echo -en "\$PSRF100,1,4800,8,1,0*0E\x0d\x0a" > /dev/ttyUSB0`

**Comment by skypuppy** [comment]
remove the gps puck and run it again.  If the 'bridge' does not show up,
there is your clue.

**Comment by iposva** [comment]
@skypuppy The comment to @cyoung was about the fact that from the list of USB devices it is not easy to distinguish whether the GPS puck is connected to /dev/ttyUSB0 or whether there is another device using the same Prolific 2303 based USB<->serial bridge.

**Comment by skypuppy** [comment]
If your system is dedicated to stratux, nothing else will BE connected
via a bridge.

**Comment by iposva** [comment]
Nothing else will be connected unless you are using the stratux in a setup which outputs to serial. See [comment](https://github.com/cyoung/stratux/pull/231#issuecomment-179456358) by @cyoung and the [serialout branch](https://github.com/cyoung/stratux/tree/serialout) branch.

**Comment by Ergonomicmike** [comment]
Just to send a "warm fuzzy," I incorporated the new ry385ai.go into my Stratux. I confirm that it automatically changed the baud rate on the BU-353-S4 from its default 4800 to 38400 and the NMEA sentences are screaming in at 5 Hz. (Would be curious to know where iposva found the cheat codes for 5 Hz. Also, since iFly thought informally thought that 5 Hz would be fine, and since @skypuppy said that fast data on the Pi USB ports is problematic & taxes the kernwl, perhaps the refresh rate for the ublox should be lowered from 10 Hz?)

**Comment by Ergonomicmike** [comment]
I got another response from the Sr. Software Engineer at iFly about refresh rates for NMEA sentences. (This is pertinent to the currently selected 10 Hz refresh rate of the ublox.)

> FYI, IMO, there is never a reason for more than 8 Hz.   It’s very simple to do “interpolation” to smooth out any animation – we don’t have to “snap” to each update, but can just lerp it, making it look like 30 FPS, by taking 4 frames to show it… this would have a 0.125 sec lag, which seems reasonable in this context.
> 
> So the ONLY thing a faster update rate does is reduce lag…   is 0.2 sec of lag too much?  0.125 sec?  0.1 sec?     this is why 15 Hz is truly unreasonable –it’s almost like they’re pretending that you’ve got to have faster than 0.1 sec lag, and that’s just not needed.

**Comment by Ergonomicmike** [comment]
When I plug my Globalsat into one of my desktop's many USB ports and then start the Windows GPSInfo program, the program knows immediately what comm port the GS has been plugged in to. If it can detect it, perhaps the Stratux can to? (Or is this because Windows passes info thru the port?) Anything that I can do that's easy to gether data?

**Comment by skypuppy** [comment]
Examine how "lsusb" code is written.

**Comment by Ergonomicmike** [comment]
Hah! I said easy! (grinning)

**Comment by Ergonomicmike** [comment]
For grins, I ran lsusb -v. Interestingly, the Edimax and the Realtek SDR's id themselves by name. But as @iposva found, instead of "GlobalSat," I'm seeing something called a Prolific USB to Serial Controller. I am guessing that that is the bridge chip inside the GlobalSat. If so, then therein lies the problem. Since the Edimax and the SDR's report who they are, I presume that GlobalSat could have done that too. But didn't. Even if they could, I doubt they would update their firmware just for us. 

So, next, just how prolific is the Prolific USB-Serial Controller in devices that the Stratux is likely to use? It would be a terrible hack, but could we assume that anything that reports as the Prolific must be the BU-353-S4?

**Comment by Ergonomicmike** [comment]
Although I'm back to wondering how Windows knows which USB port the GlobalSat is hooked up to. There must be some data being sent by the GlobalSat. (Unless GPSInfo assumes that the Prolific must be the puck.) I went thru Windows' equivalent of lsusb in Device Manager. All about Prolific. Nothing about GlobalSat.

**Comment by jzeevi** [comment]
Using the prolific info is the wrong approach. With lsusb -v you should be able to see the sub-pid and sub-vid which hopefully identify the globalsat part. The Prolific part is, er, well prolific. 

Josef

**Comment by Ergonomicmike** [comment]
I don't see a sub-pid explicitly. Here's a screen dump:


```
Bus 001 Device 006: ID 067b:2303 Prolific Technology, Inc. PL2303 Serial Port
Device Descriptor:
  bLength                18
  bDescriptorType         1
  bcdUSB               1.10
  bDeviceClass            0 (Defined at Interface level)
  bDeviceSubClass         0
  bDeviceProtocol         0
  bMaxPacketSize0        64
  idVendor           0x067b Prolific Technology, Inc.
  idProduct          0x2303 PL2303 Serial Port
  bcdDevice            4.00
  iManufacturer           1 Prolific Technology Inc.
  iProduct                2 USB-Serial Controller D
  iSerial                 0
  bNumConfigurations      1
  Configuration Descriptor:
    bLength                 9
    bDescriptorType         2
    wTotalLength           39
    bNumInterfaces          1
    bConfigurationValue     1
    iConfiguration          0
    bmAttributes         0x80
      (Bus Powered)
    MaxPower              100mA
    Interface Descriptor:
      bLength                 9
      bDescriptorType         4
      bInterfaceNumber        0
      bAlternateSetting       0
      bNumEndpoints           3
      bInterfaceClass       255 Vendor Specific Class
      bInterfaceSubClass      0
      bInterfaceProtocol      0
      iInterface              0
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x81  EP 1 IN
        bmAttributes            3
          Transfer Type            Interrupt
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x000a  1x 10 bytes
        bInterval               1
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x02  EP 2 OUT
        bmAttributes            2
          Transfer Type            Bulk
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0040  1x 64 bytes
        bInterval               0
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x83  EP 3 IN
        bmAttributes            2
          Transfer Type            Bulk
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0040  1x 64 bytes
        bInterval               0
Device Status:     0x0000
  (Bus Powered)
```

**Comment by jzeevi** [comment]
Then a query of some sort is likely required of the device connected as this serial port. :-(
Bummer. 

Josef

**Comment by jzeevi** [comment]
Oh- can you try plugging it in to Windows and look through the properties for more info? There's usually a host of things that may be of help. 

Josef

**Comment by Ergonomicmike** [comment]
It does respond to commands, somewhat like old Modems used to.

**Comment by JohnOCFII** [comment]
> So, next, just how prolific is the Prolific USB-Serial Controller in devices that the Stratux is likely to use?

Sadly,it is very prolific.  I have one in at least 3 devices that were sold as "USB" devices.  All three of those are commonly used with Linux systems (Not necessarily in _my_ Stratux, but for folks who use their Pi for multiple functions, this could be an issue.)

**Comment by Ergonomicmike** [comment]
From the [GlobalSat FAQ](http://usglobalsat.com/store/gpsfacts/bu353s4_gps_facts.html), 

```
31Q) How do I check the Firmware version?

31A) You can check the firmware by sending a cold start or initialization command to the GPS Module.

Send this command:

$PSRF101,0,0,0,0,0,0,12,4*10

You should receive something similar to this:

$PSRFTXT,Version:GSW3.5.0_3.5.00.00-SDK-3EP2.01 *46

$PSRFTXT,Version2:F-GPS-03-0904011*20
```
I see the 3 key characters "GPS" in the string. Can a smart guy use that?

Ergonomicmike:
I just tried it. It reports the baud rate and keeps it the same as it was. But it does do a "cold start" of sorts, and that makes the puck wait about 10 seconds before it blinks again. I can't tell if the refresh rate dropped back to 1 Hz. But presumably, this sniffing would be done first, and then baud rate and refresh rate would be set by Stratux until the next power down restart.

skypuppy:
Prolific is not the only one out there, but it is the best design, so far.

skypuppy:
Also, don't your SDR's connect directly into a USB port?  The bridge
sits, electrically and physically, between the USB hardware and whatever
it is connected to, the GPS in our case.

Sadly, GPS is about the only thing (at least in my world) that still
uses old RS-232 serial.  <sigh>

skypuppy:
Sometimes one can query a device through the bridge and get the needed
info.  Case by case basis, though.

skypuppy:
The vendor or manufacturer is the place to find out those kinds of
specifics.

PR #244, opened 2016-02-12
Additional GPS compatibility improvements

AvSquirrel:
1. Cleaned up groundspeed detection routines to improve consistency of reporting. Groundspeed will now be reported any time a valid speed is parsed. True course will be updated only when groundspeed > 3 knots. Method is now consistently applied to all source message types (PUBX,00, VTG, and RMC).

2. Add missing `LastFixLocalTime` and `LastGroundFixTime` updates to VTG and RMC parsers to improve compatibility with older / generic NMEA receivers. A minimum of RMC + GSA are all that's needed to generate a valid 2D fix, including speed / track. 3D position still requires GGA messages.

3. Quality and position accuracy are now parsed prior to position reports, rather than in NMEA sentence order. This allows values to be updated prior to returning on a bad / missing position, and ensures that invalid lat/lon are not returned from PUBX messages. (Bad position messages didn't get passed to the GDL-90 output, but would show up on the web UI and JSON interfaces.)

### DIFF
--- a/main/ry835ai.go
+++ b/main/ry835ai.go
@@ -812,6 +812,8 @@ func processNMEALine(l string) bool {
 		if x[2] != "A" { // invalid fix
 			mySituation.quality = 0
 			return false
+		} else if mySituation.quality == 0 {
+			mySituation.quality = 1 // fallback option; indicate if the position fix is valid even if GGA or PUBX,00 aren't received
 		}
 
 		// Timestamp.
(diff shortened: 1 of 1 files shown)


main/ry835ai.go
```diff
@@ -408,6 +408,17 @@ func calculateNACp(accuracy float32) uint8 {
 	return ret
 }
 
+/*
+processNMEALine parses NMEA-0183 formatted strings against several message types.
+
+Standard messages supported: RMC GGA VTG GSA
+U-blox proprietary messages: PUBX,00 PUBX,03 PUBX,04
+
+return is false if errors occur during parse, or if GPS position is invalid
```
cyoung: What about the case when there is no position in the line being parsed? Is that an "invalid position"?

AvSquirrel:
@cyoung -- yes for GGA and RMC messages. Missing or malformed lat / lng fields will have fewer than four or five digits respectively, and therefore [return false](https://github.com/cyoung/stratux/blob/master/main/ry835ai.go#L733-L735).

For the PUBX messages, I'm relying on the [solution type message](https://github.com/cyoung/stratux/blob/master/main/ry835ai.go#L449-L482) to do the validation and return early, since PUBX,00 always passes raw lat/lng -- e.g at startup, it sends 0°N, 0°E. Time will be set prior to position lock by using the PUBX,04 message; time validation is performed by checking that the GPS has a valid GPS week.

Issue #79, opened 2015-10-10
Support two PPM if there are two SDRs


bradanlane:
I've run the Kalibrate utility against the two SDRs in my setup. I have measurably different results for the two SDRs.

If others are seeing a similar result then should we allow users to set PPM for each SRR?

I suspect we only set it in the first since that is the one used for UAT and thus more critical - correct?

If we feel PPM will not provide an improvement on the 1090ES receive, then the resolution is to improve the help text for PPM to explain this to the user.

224XS:
I have two different model r820T sdrs, one older, the other the new version. Both seem to produce the same identifying data tags.  Is there a way to bind the respective frequency offsets to their proper SDR?

bradanlane:
It's actually worse that that. I've tested older R802T units and newer R802T2 units and they all report the exact same signatures.

ssokol:
The best option for this is probably to use the physical port, mapped
through udev. Here's a post on it:

http://askubuntu.com/questions/49910/how-to-distinguish-between-identical-usb-to-serial-adapters

On Sat, Oct 10, 2015 at 4:59 AM, bradanlane <notifications@github.com>
wrote:

> It's actually worse that that. I've tested older R802T units and newer
> R802T2 units and they all report the exact same signatures.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/cyoung/stratux/issues/79#issuecomment-147070599>.
>



-- 
Steven Sokol
408 Camelot Drive
Liberty, MO 64068

mobile: +1 816-806-8844
fax: +1 816-817-0441

tbirdrv:
Has anyone looked at [rtl-eeprom](http://manpages.ubuntu.com/manpages/trusty/man1/rtl_eeprom.1.html)? 

Would it be possible to encode a unique name or ID to each SDR? I'm not even sure this program works with our SDRs.

jpoirier:
Not sure if it's the right approach but rtl-eeprom is in librtlsdr
therefore it's available in gortlsdr via WriteEeprom.

cyoung:
https://github.com/osmocom/rtl-sdr/blob/master/src/rtl_eeprom.c

@jpoirier - do you think a "set/get serial string" function belongs in gortlsdr?

jpoirier:
@cyoung I hadn't looked at the rtl_eeprom.c file until just now. Having
similar functionality in gortlsdr would be useful. I can update the package
later this evening unless you or someone else wants to start on it now.

[...]

PR #248, opened 2016-02-13
Fix for Issue #79

yihchun:
Allows for per-SDR PPM information.

- Legacy support for stratux:978 and stratux:1090 is preserved
- SDRs without a specified PPM get the globalSettings PPM as configured on the Settings page
- Because the serial number field is length-constrained, the preferred format with PPM is stx:freq:ppm, such as stx:1090:20. Per peepsnet's question below, negative PPMs are supported; example: stx:1090:-20
- The PPM parser and the frequency designator are independent entities. If you don't need to specify a frequency, pick any number other than 978 and 1090 (0 recommended), and Stratux will assign frequencies. (Per peepsnet's second question below). Example: stx:0:-20

### DIFF
--- a/main/sdr.go
+++ b/main/sdr.go
@@ -13,11 +13,12 @@ import (
 	"io"
 	"log"
 	"os/exec"
+	"regexp"
 	"strconv"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
-	"regexp"
 
 	"../godump978"
 	rtl "github.com/jpoirier/gortlsdr"
@@ -127,8 +128,12 @@ func (u *UAT) read() {
 }
 
 func getPPM(serial string) int {
-	r, _ := regexp.Compile("str?a?t?u?x:\\d+:?(-?\\d*)")
-        arr := r.FindStringSubmatch(serial)
+	r, err := regexp.Compile("str?a?t?u?x:\\d+:?(-?\\d*)"); 
+	if err != nil {
+		return globalSettings.PPM
+	}
+
+	arr := r.FindStringSubmatch(serial); 
 	if arr == nil {
 		return globalSettings.PPM
 	}
@@ -316,6 +321,19 @@ func sdrKill() {
 
 // Watch for config/device changes.
 func sdrWatcher() {
+	var doSkip bool
+	rES, err := regexp.Compile("str?a?t?u?x:1090")
+	if err != nil {
+		rES = nil
+		log.Println("failed to compile ES regexp because %s", err.Error())
+	}
+
+	rUAT, err := regexp.Compile("str?a?t?u?x:978")
+	if err != nil {
+		rUAT = nil
+		log.Println("failed to compile ES regexp because %s", err.Error())
+	}
+
 	for {
 		time.Sleep(1 * time.Second)
 		if sdrShutdown {
@@ -385,8 +403,13 @@ func sdrWatcher() {
 					serial = ""
 				}
 
-				r, _ := regexp.Compile("str?a?t?u?x:1090")
-				if !r.MatchString(serial) {
+				if (rES == nil) {
+					doSkip = rES.MatchString(serial)
+				} else {
+					doSkip = strings.Compare(serial, "stratux:1090") == 0
+				}
+
+				if !doSkip {
 					UATDev = &UAT{indexID: id, serial: serial}
 					if err := UATDev.sdrConfig(); err != nil {
 						log.Printf("UATDev = &UAT{indexID: id} failed: %s\n", err)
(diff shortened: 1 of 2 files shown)


peepsnet:
Will this work for negative ppm vals??  stx:978:-25

yihchun:
Yes, the regexps include a -? to capture possibly negative numbers.

peepsnet:
Great! Im sold!!

peepsnet:
Another question. There was a post made to address a SDR failure in flight. If the 978 SDR failed in flight and your other SDR's serial was stratux:1090 then you would never be able to get weather after a reboot because the working SDR would tune to 1090.  The Solution was to set only one SDR serial to stratux:978 and the other to something else(stratux:atl suggested) This would allow the this SDR to not be forced to 1090 if detected first. 

Does your code allow for this?  
SDR 1: stx:978:-24
SDR 2: stx:alt:-27

Would the code set SDR 2 to 978 if there is only one SDR detected but 978 & 1090 turned on in the settings??

Does this make sense??

yihchun:
Doesn't allow for your format, but stx:0:-27 should work.

peepsnet:
ok. just something to allow the functionality.  So if there is no 978 SDR found but 978 is wanted(turned on in settings) then the remaining SDR will be set to 978 if the serial is not 1090.

yihchun:
Right. In fact, if you don't have frequency-specific antennas, you can set both of your radios to stx:0:-24 and stx:0:-27 and Stratux will always decide for you.

peepsnet:
What freq is set first?  is it 978 by default??

yihchun:
If you only have one, it'll shut down the 1090 to switch it to 978, unless 1090 is explicitly set in serial.
"When count is one, favor UAT in the case where the user has enabled both UAT and ES via the web interface."

peepsnet:
That is the Correct logic!  I believe that is a favored logical path for a failure. You would want weather more then traffic...  Great job on the code. That is the idea from the piaware software, correct? I believe I saw this on an internet search. Glad it can be added here!!

peepsnet:
I think the next step for this is to have a check box in the settings webui which allows the user to determine if they want to override the 1090 SDR if the 978 fails. This would allow the user to set the serials to match the SDR's and not lose the prioritization

yihchun:
The code doesn't borrow from piaware or anything; it's just a simple regexp matching.

As for webui / default behavior changes, that's out of the scope of this pull request, which is limited to changing the format of the serial number, and integrating the PPM therein. It's also beyond the scope of Issue #79 .

peepsnet:
I know that is beyond this pull... But to allow anybody access to this functionality it should be added to settings and allowed to choose priority. But in order to get the PPM offset I guess one would need some knowledge. Enough to not need a web interface. 

But maybe someday it can all be in the webUI... Even the kal functionality...

main/sdr.go
```diff
@@ -365,7 +384,9 @@ func sdrWatcher() {
 				if err != nil {
 					serial = ""
 				}
-				if strings.Compare(serial, "stratux:1090") != 0 {
+
+				r, _ := regexp.Compile("str?a?t?u?x:1090")
```
jpoirier: The regexp string isn't changing so no reason to compile it with each pass, just do it once on entry in to the function prior to dropping in to the for loop. Why no error check/handling?

main/sdr.go
```diff
@@ -407,7 +428,8 @@ func sdrWatcher() {
 				if err != nil {
 					serial = ""
 				}
-				if strings.Compare(serial, "stratux:978") != 0 {
+				r, _ := regexp.Compile("str?a?t?u?x:978")
```
jpoirier: ...same comment as for r, _ := regexp.Compile("str?a?t?u?x:1090")

main/sdr.go
```diff
@@ -124,12 +126,29 @@ func (u *UAT) read() {
 	}
 }
 
+func getPPM(serial string) int {
+	r, _ := regexp.Compile("str?a?t?u?x:\\d+:?(-?\\d*)")
```
jpoirier: Why no error check/handling?

yihchun:
Thanks for the comments. I didn't error check the regexp compiles because I figured the regexps are static, so they should compile OK. I've put in error checking code now, with a fallback to string.Compare.

jpoirier:
LGTM but the merge is Chris' call

cyoung:
Thanks @yihchun, and thanks @jpoirier for reviewing.

cyoung:
@yihchun - need to set up your github client so that your commits aren't showing up as "root"

yihchun:
Will do, sorry about that.

peepsnet:
Is the PPM setting on the Settings page still used at all? And if so when?
And did you make any log entries for setting the ppm?
And what was the format? stx:xxx:ppm or str:xxx:ppm? The ereg starts str?a?

It apprears the code behind the ppm field in settings is broken. If I delete the text it returns after leaving the page and returning.

yihchun:
PPM setting is used when no explicit per-dongle PPM is set.

The log entries for UAT PPM remains unchanged; there was no prior log message for ES PPM, and I did not add one.

The format is stx:number:ppm. ? means that the previous element is optional.

I didn't touch the settings field.

Issue #246, opened 2016-02-13
Wi-Fi does not reconnect when Edimax unplugged/plugged


Ergonomicmike:
Some users (including me) report random (seemingly) Wi-Fi disconnects. We've been [brainstorming on the reddit](https://www.reddit.com/r/stratux/comments/45bqox/help_catchingreproducing_wifi_disconnects/) about what could be causing it.

I don't know if what I found here is ultimately responsible for the disconnect. But if you unplug and then replug the Edimax, the Wi-Fi connection never comes back.

Now, it seems to me that, just like when the GPS is unplugged and replugged, the Edimax should gracefully terminate when unplugged and then restart when re-plugged. At least then we could eliminate a small scale physical disconnect (perhaps by a low power transient or an intermittent connection at a micro-level?) causing a large scale permanent disconnect.

Here's an excerpt from syslog when I unplugged the Edimax:

```
Feb 13 00:25:57 raspberrypi dhcpd: receive_packet failed on wlan0: Network is down
Feb 13 00:25:57 raspberrypi kernel: [  178.607381] usb 1-1.2: USB disconnect, device number 4
Feb 13 00:25:57 raspberrypi kernel: [  178.607972] rtw_cmd_thread: DriverStopped(0) SurpriseRemoved(1) break at line 482
Feb 13 00:25:58 raspberrypi ifplugd(wlan0)[1601]: Exiting.
```

And here's an excerpt when I plugged the Edimax back in. (It never reconnects.)

```
Feb 13 00:27:01 raspberrypi kernel: [  241.826018] usb 1-1.2: new high-speed USB device number 8 using dwc_otg
Feb 13 00:27:01 raspberrypi kernel: [  241.928046] usb 1-1.2: New USB device found, idVendor=7392, idProduct=7811
Feb 13 00:27:01 raspberrypi kernel: [  241.928079] usb 1-1.2: New USB device strings: Mfr=1, Product=2, SerialNumber=3
Feb 13 00:27:01 raspberrypi kernel: [  241.928097] usb 1-1.2: Product: 802.11n WLAN Adapter
Feb 13 00:27:01 raspberrypi kernel: [  241.928113] usb 1-1.2: Manufacturer: Realtek
Feb 13 00:27:01 raspberrypi kernel: [  241.928129] usb 1-1.2: SerialNumber: 00e04c000001
Feb 13 00:27:01 raspberrypi ifplugd(wlan0)[2288]: ifplugd 0.28 initializing.
[...]

PR #249, opened 2016-02-15
Networking improvements

AvSquirrel:
These changes improve network message efficiency, and are a possible solution to some of the disconnect problems described in #246.

Testing with both UAT replay and simulated traffic data indicates that wlan0 does not handle high IO rates well. Details of this testing will be posted to #246. Although I didn't determine whether this was a go issue, a wlan driver issue, or a hardware / power issue, I did find opportunities to optimize both the queued and nonqueued message senders.

- Traffic data is now passed to `sendGDL90()` in a single message for each call to sendTrafficUpdates(), rather than as sequential individual reports. This has no effect on timing of traffic data, but greatly improves IO efficiency. With two ForeFlight clients connected, I found that the wlan connection would drop / disconnect after ten minutes with 100 traffic targets, after two minutes with 250 targets, and after 15 seconds with 1000 targets (the last with a message rate of over 2000 messages/sec). With the traffic reports "batched", I was able to run for over five hours, with typical message rates of 32 IOPS.
- The message queue in `network.go` now batches 256 FIS-B messages before sending them to clients. This will improve stability when bringing devices out of standby; during one replay test before this was implemented, I observed rates of 800+ messages/sec as the queue emptied.
- The ForeFlight AHRS sender was commented out to save 10 messages/sec per channel.
- Fixed an issue where messages were not queuing for sleeping devices
- Added log status messages for message queue depth per IP address and port
- Added log status messages for IO messages / bytes; detailed (per second) status can be had by uncommenting network.go L422
- Added synthetic traffic routine to add an arbitrary number of targets to the traffic map for debug / testing purposes. The call in gen_gdl90.go is commented out in this PR, but can be enabled / compiled as needed by developers.

### DIFF
--- a/main/network.go
+++ b/main/network.go
@@ -150,14 +150,22 @@ func sendToAllConnectedClients(msg networkMessage) {
 					netconn.messageQueue = netconn.messageQueue[s:]
 				}
 			}
-			netconn.messageQueue = append(netconn.messageQueue, msg.msg)
+			netconn.messageQueue = append(netconn.messageQueue, msg.msg) // each netconn.messageQueue is therefore an array (well, a slice) of formatted GDL90 messages
 			outSockets[k] = netconn
 		}
 	}
 }
 
-// Just returns the number of DHCP leases for now.
+// Returns the number of DHCP leases and prints queue lengths
 func getNetworkStats() uint {
+	for _, netconn := range outSockets {
+		queueBytes := 0
+		for _, msg := range netconn.messageQueue {
+			queueBytes += len(msg)
+		}
+		log.Printf("On  %s:%d,  Queue length = %d messages / %d bytes\n", netconn.Ip, netconn.Port, len(netconn.messageQueue), queueBytes)
+	}
+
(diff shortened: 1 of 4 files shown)


braddward:
Wow - great work.  Has anyone also considered doing some tuning on the tcp
sysctl parameters?

cyoung:
@braddward https://github.com/cyoung/stratux/commit/5aca8adea2b4f8c6fe7baf312088bafa152a73be

main/network.go
```diff
@@ -124,14 +124,20 @@ func sendToAllConnectedClients(msg networkMessage) {
 		}
 		// Send non-queueable messages immediately, or discard if the client is in sleep mode.
 
-		if sleepFlag {
-			continue
```
cyoung: Good catch.

cyoung:
Excellent work, I'll merge it now but have concerns on app compatibility. We'll have to circle back to evaluate whether the change in packet format confuses any EFBs.

Issue #254, opened 2016-02-17
Stratux WebUI Stops Showing Traffic


pp4life:
1. Stratux version: 0.7b1
2. Stratux config: dual SDR, GPS no, AHRS no, battery Anker, battery cable, Anker Standard
3. EFB app and version Foreflight 7.5.2. EFB platform (iOS 8.3 on iPad Mini 2, iOS 9.1 on iPod Touch 5th Generation). EFB hardware iPad Mini 2, iPod Touch 5th Gen.
4. Description of your issue.
See linked video.
https://youtu.be/8MTvH-tOQk4
       Ignore the Skywest target in the upper left of Foreflight. It got stuck there, I don't know why, but it was still there after the lower right target went out of range. Took a restart of Foreflight for it to go away.
        All webUI stuff is done on the iPod Touch, and all Foreflight stuff is done on the iPad mini. But the the general problem is that the traffic page on the Stratux WebUI stops showing traffic, but Foreflight will continue to show the traffic.
       If I tap menu on the WebUI and switch over to the Status page then flip back to the Traffic page it will again show the traffic that I'm seeing in Foreflight. I would have never noticed without having two devices sitting right next to each other so I could test for issue #250. It's like the Traffic page stops updating.

AvSquirrel:
Noticed this last night while debugging #251.

Every minute, [this bit of Javascript](https://github.com/cyoung/stratux/blob/master/web/plates/js/traffic.js#L110-L126) looks for any traffic entries that haven't been refreshed in more than three minutes. Problem is that JS uses runs on the client (i.e. iPad), and therefore uses that device's time / date. Traffic time is based on an uptime counter that starts at 0, and therefore times out all traffic each time it's called.

This doesn't affect the targets send to ForeFlight, and traffic does come back on the UI after a second or two, but it should (and will) be fixed.

Ergonomicmike:
In the comment in the java script code above it says "stail aircraft." I politely ask if that's supposed to be "stale"?

bradanlane:
there are a few things happening here. First, the web ui is attempting cleanup after 180 seconds whereas the stratux system code is every 60 seconds. this is an easy fix.

the bigger challenge is the definition of "time". this one keeps biting stratux. there are several definitions of time:

"Stratux Time" - the clock time on the Raspberry Pi, which is undefined if we don't have a gps - it could be just about anything
"Data Time" - the time stamped by the FAA
[...]

Issue #251, opened 2016-02-16
Timestamp on Traffic WebUI row seems not right


bkwny:
1. Stratux version:
6b1 plus the first 29 commits since it.

2. Stratux config: single/dual SDR, GPS yes/no (type?), AHRS yes/no, battery, battery cable.
Dual SDR Nano 2.  RY835AI via GPIO.  AHRS is disabled.  Line power via Turnigy 5.1V 3A battery eliminator powered by 12 VDC.  From "tail -f /var/log/stratux.log" I can see the R Pi system time is set to UTC.

3. EFB app and version (e.g., WingX Pro7 8.6.2). EFB platform (e.g., iOS 9.2). EFB hardware (e.g., iPad Mini 2).
N/A for this issue.

4. Description of your issue. If possible, enable "Replay Logs", reproduce the problem, and provide a copy of the logs in http://192.168.10.1/logs/stratux/ and http://192.168.10.1/logs/stratux.log.
On the WebUI traffic display, for each traffic row, the last column seems to be a time stamp of some kind, for example, 04:31:38Z and increasing.  But the value printed is not a UTC timestamp nor is it a local timestamp.  There is some speculation that maybe it is the hh:mm:ss amount of time that has transpired since the R Pi was booted.

cyoung:
Please use the issue template: https://github.com/cyoung/stratux/blob/master/CONTRIBUTING.md (linked at the top of the "New issue" page).

Some relevant information needed: version you're running and GPS or not (since the clock of the RPi is set off of the GPS when connected, not set at all if not connected).

Ergonomicmike:
Not a big deal to me, but I've got the GlobalSat GPS. Internal Pi time synchs (per date command) to GPS. But the Traffic time stamp does not. My time stamps reset to 00:00Z on each reboot. Did it on v0.6r1 and still on v0.7b1.

AvSquirrel:
As noted on reddit, this is time since boot, and is on my to-do list.

Stratux has used time since boot over the last few releases to track all time-critical activities. These include timing out old traffic, determining if GPS fix is recent, sending log messages, etc. This approach is used because the RPi lacks a realtime clock, and "jumping" the time forward or backward (as might happen with a GPS or UAT uplink providing time data) has caused odd problems in the past.

cyoung:
We're not saving the incoming timestamps on traffic reports. Would you just use those?

AvSquirrel:
Well, there are two ways I could go about this.

[...]

PR #259, opened 2016-02-18
Traffic UI and other improvements

AvSquirrel:
Fixes for #251 and #254.

- Traffic "last seen" time is now changed to a traffic age, set by the internal Stratux `monotonic` timer.
- traffic.go JSON calls updated to refresh all traffic once per second, even if postion hasn't been updated.
- New ti.Timestamp and ti.Age variables added to traffic.go
- Timeout for sending GDL90 traffic message reduced to 6 seconds. Traffic will remain in internal database and on web UI for 60 seconds.
- Minor GPS improvements related to time parsing
- Added clock to web UI status page.
- Update status UI to reflect new features added over the past few months


cyoung:
Nice work.

PR #261, opened 2016-02-18
Fixed a bug from #248

yihchun:
When I added error checking on the regexps I introduced a bug due to incorrect logic. Sorry about that.


jpoirier:
NP, it happens. And I should have been more diligent reviewing the changes. :( Thanks for getting to this quickly.

PR #264, opened 2016-02-18
Bug fixes for traffic UI and setting GPS time

AvSquirrel:
- Fix a bug in #259 where traffic targets 59.x seconds old are not removed from web UI
- Fix a malformed time string when setting date / time from the RMC GPS message


skypuppy:
AvSquirrel, maybe it would serve better if you used >= 60.0  (or 59.0)
rather than just =?

PR #266, opened 2016-02-18
use standalone httpd service / move stratux services to port 8080

bradanlane:
this removes gen_gdl90 from service the basic WebUI. gen_gdl90 still provides its web services and websockets - now on port 8080.

`http://192.168.10.1` will return the Stratus WebUI
`http://192.168.10.1:8080/getSettings` will return the stratux settings JSON

Spindle will install `lighttpd`, add a symlink to access logs, and then enable support for directory lists and smlinks so the log page still works.

### DIFF
--- a/image/spindle/wheezy-stage4
+++ b/image/spindle/wheezy-stage4
@@ -37,8 +37,14 @@ setup_stratux() {
 echo "**** STRATUX SETUP *****"
 
   ssh_in_to_qemu chroot /mnt sh -l -ex - <<\EOF
-#general
-apt-get install -y screen
+
+#http server
+apt-get install -y lighttpd
+ln -s /var/log /var/www/logs
+makedir -p /var/logs/stratux
+chmod o+x /var/logs/stratux
+echo 'dir-listing.activate = "enable"' >> /etc/lighttpd/lighttpd.conf
+echo 'server.follow-symlink = "enable"' >> /etc/lighttpd/lighttpd.conf
 #wifi
 apt-get install -y hostapd isc-dhcp-server
 #troubleshooting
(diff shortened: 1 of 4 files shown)


cyoung:
I'll defer merging until the last minute before making another image - since to transition we'll all need to install a new image. Everything working in the web interface with the new port?

bradanlane:
understood.

I tested each of the webui pages to insure they still connected to stratux using the new port. I also manually tested the web service interfaces with their direct URLs. I did the necessary config for lighttpd to support the logs directory and stratux.log file.

cyoung:
Installed it, set it up, started working on it and a bunch of CORS issues were coming up in different browsers. The Go webserver works great, has since the start, never had a single problem. We'll stick with that until there's some material benefit we can gain from switching.

bradanlane:
"CORS" issues? I'm not familiar with that term. 

I tested with Chrom on desktop as well as Chrome on Android and Safari Mobile on iPhone and iPad. 

There is a required "permissions" setting on the directory. If that is missed, a bunch of stuff gets blocked.

cyoung:
http://stackoverflow.com/questions/19743396/cors-cannot-use-wildcard-in-access-control-allow-origin-when-credentials-flag-i

The permissions were fine, the webserver was serving the necessary files just fine. The issue was in the browser and the origin being http://192.168.10.1 but accessing resources on http://192.168.10.1:8080/.

```
XMLHttpRequest cannot load http://192.168.10.1:8080/updateUpload. No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://192.168.10.1' is therefore not allowed access.
```
etc

bradanlane:
Gotcha. I understand you will not merge it but can you tell me what browser and OS you were using?

cyoung:
OS X El Capitan, Chrome 48.0.2564.116

bradanlane:
Thanks. I don't have a Mac so I won't be able to thatcthey OS but I'll try that version of Chrome

PR #268, opened 2016-02-19
Show traffic source (ADS-B vs TIS-B) in web UI

AvSquirrel:
Requested at https://redd.it/46hmso

This update passes `ti.Addr_type` to the web UI to support showing whether traffic messages originated from ADS-B air-to-air broadcasts, or from TIS-B ground-to-air broadcasts. Currently, this is supported only for UAT traffic, as the dump1090 output on port 30003 does not pass this status. 

Traffic reports originating at the GBTs are indicated with a lighter highlight color and a satellite dish symbol. Airborne reports use the existing color and get an airplane symbol.

image

### DIFF
--- a/main/traffic.go
+++ b/main/traffic.go
@@ -67,7 +67,7 @@ const (
 type TrafficInfo struct {
 	Icao_addr        uint32
 	OnGround         bool
-	addr_type        uint8
+	Addr_type        uint8
 	emitter_category uint8
 
 	Lat float32
@@ -140,7 +140,7 @@ func makeTrafficReportMsg(ti TrafficInfo) []byte {
 	// See p.16.
 	msg[0] = 0x14 // Message type "Traffic Report".
 
-	msg[1] = 0x10 | ti.addr_type // Alert status, address type.
+	msg[1] = 0x10 | ti.Addr_type // Alert status, address type.
 
 	// ICAO Address.
 	msg[2] = byte((ti.Icao_addr & 0x00FF0000) >> 16)
@@ -241,10 +241,10 @@ func parseDownlinkReport(s string) {
 	}
 	ti.Icao_addr = icao_addr
 
-	ti.addr_type = uint8(frame[0]) & 0x07
+	ti.Addr_type = uint8(frame[0]) & 0x07
 
 	// OK.
-	//	fmt.Printf("%d, %d, %06X\n", msg_type, ti.addr_type, ti.Icao_addr)
+	//	fmt.Printf("%d, %d, %06X\n", msg_type, ti.Addr_type, ti.Icao_addr)
 
 	nic := uint8(frame[11]) & 15 //TODO: Meaning?
 
@@ -568,7 +568,7 @@ func esListen() {
 			ti.Last_seen = stratuxClock.Time
 			//ti.Age = 0
 
-			ti.addr_type = 0           //FIXME: ADS-B with ICAO address. Not recognized by ForeFlight.
+			ti.Addr_type = 0           //FIXME: ADS-B with ICAO address. Not recognized by ForeFlight.
 			ti.emitter_category = 0x01 //FIXME. "Light"
 
 			// This is a hack to show the source of the traffic in ForeFlight.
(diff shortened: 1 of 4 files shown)


bkwny:
So, check me on this, please.  There are four possible visuals in the first column, meaning these things:

1. Dark tan with airplane icon:  air-to-air 978
2.  Yellow with satellite dish icon:  ground-to-air 978, not distinguishing between a TIS-B packet and an ADS-R packet
3.  Dark blue with airplane icon:  air-to-air 1090
4.  Light blue with satellite dish icon:  ground-to-air 1090, not distinguishing between a TIS-B packet and an ADS-R packet

Is that right?

AvSquirrel:
@bkwny - that's correct, again with the caveat that the current build of dump1090 doesn't tell us the source of 1090 traffic (all 1090 targets are assumed to be air-to-air ADS-B transmissions)

Technical details ahead.... Because of the way the UAT messages are structured, it isn't not possible differentiate between TIS-B and ADS-R targets if the aircraft in question is transmitting a Mode S address. This comes down to how the "address qualifier" field in the first byte of the UAT traffic message is assigned:

- 0 indicates ADS-B target with a 24-bit ICAO address (transmitted by aircraft)
- 1 is reserved
- **2 indicates TIS-B or ADS-R target with a 24-bit ICAO address**
- 3 indicates TIS-B target with a track-file identifier assigned by the ground station
- 4 indicates ground vehicle
- 5 indicates fixed ADS-B beacon
- 6 indicates ADS-R target with a non-ICAO address

Example: Let's say I have a Navworx UAT box, and it's telling the GBT that I'm a 978 In client. There's three other targets inside my "hockey puck": a C172 with an old Narco transponder,  a C210 transmitting 1090 ADS-B Out from a 430W+330ES combo, and a CRJ-700 that has a Mode S transponder for TCAS, but hasn't been upgraded to ADS-B yet.

The ground station will three traffic targets for me: a TIS-B message with address qualifier = 3 for the C172, an ADS-R message with address qualifier = 2 for the C210, and a TIS-B message with address qualifier = 2 for the CRJ.

I know for certainty that the "3" qualifier for the C172 is TIS-B, but address qualifier 2 is indeterminate.

skypuppy:
Not trying to be a smart-alek, but is it important to know which of the
bit 2 source it is?  I must have lost the context here...

Skypuppy

bkwny:
@AvSquirrel, thank you for your work.  What do you think about this:  until dump1090 can tell you the difference between ground-sourced packets and air-sourced packets, for the colored rectangle, draw a blue rectangle with a black question mark for the icon?  When dump1090 gets changed, then begin using the satellite-dish and airplane icons you have made.  Thank you again for your work.

cyoung:
What is the difference between airplane+light tan vs. airplane+dark tan?

http://i.imgur.com/VJAARN8.png

AvSquirrel:
Light blue / light tan indicates that the traffic broadcast was received from a ground uplink, rather than air-to-air transmission. Airplane icon indicates ADS-R rebroadcast of traffic on the "other" frequency. Antenna icon indicates TIS-B (position derived from radar data).

This is documented in my help file updates included with PR#340.

image

cyoung:
Still confused. I see three different colors of tan/gold for targets with an airplane:

http://i.imgur.com/VJAARN8.png

ASH6182 is is a light tan airplane icon. So it's ADS-R. Sounds right.
N102JV is a dark tan airplane icon. So it's air-to-air UAT. Also sounds right.

AA9F82 is a dark tan airplane icon with a different shade of font. What does this mean? To my eye it also looks like a shade that is somewhere between "light tan" and "gold" because of the font color. Well, it did at first until paying close attention.

Maybe we could just reduce the number of possibilities here, there's no possible way anyone will remember all of these colors and icon combinations. It looks like there are 6 possibilities of color + icon from your screenshot, plus 2 font colors which makes for 16 possibilities. Perhaps we could just use the icon to display the source of the transmitted data, as received by stratux.

AvSquirrel:
Both light tan boxes are RGB = 240,230,140. The apparent difference in color is an optical illusion due to the gray ICAO font color.

I wouldn't have a problem getting rid of the gray font. The only thing it's used for in #340 (see also the comments toward the end of #294) is the "[unknown]" callsign when callsign is not available. Just get rid of `class="text-muted"` in [your branch](https://github.com/cyoung/stratux/blob/master/web/plates/traffic.html#L29) or in [mine](https://github.com/AvSquirrel/stratux/blob/gpsattitude/web/plates/traffic.html#L32)

PR #273, opened 2016-02-20
Second wifi dongle settings template for Dev work

peepsnet:
Added settings to /etc/network/interfaces as a template for a second wifi dongle for dev work


cyoung:
Thanks, Don!

Issue #275, opened 2016-02-21
1090 sdr not initiating with new per sdr ppm serial format 


peepsnet:
I'm running a dual sdr stratux with ry835ai and edimax
Version . 7b2

I believe the issue is related to the new dual SDR PPM logic that was introduced in the last release

When booting the device the 1090 radio never initiates although it shows active on the Status page. I only get 978 data. 

Serials:
stx:978:-24
stx:alt:-28

Here's a video of the walkthrough for debugging purposes
https://drive.google.com/file/d/0B8TVybIHchkfYWZBMHRGUFh6UEk/view?usp=docslist_api

On a side note.
Can we get a bit more info on the Status page regarding the sdr's. I would like to know if the sdr is responding. In the above issue the Status page shows the sdr is just not receiving any signal not that the sdr is not even in use by stratux!  It would be nice to show an sdr is selected but not functioning

@yihchun

yihchun:
In #248 I noted
"If you don't need to specify a frequency, pick any number other than 978 and 1090 (0 recommended)" and "Doesn't allow for your format, but stx:0:-27 should work."

This will allow for better debugging:
1. After boot, run `ps auxww | grep dump1090`
2. Insert the debugging lines 
`log.Printf("===== UAT Device serial: %s =====\n", u.serial)` at the beginning of `func (u *UAT) sdrConfig() (err error)` and
`log.Printf("===== ES Device serial: %s =====\n", e.serial)` at the beginning of `func (e *ES) sdrConfig() (err error)`. Recompile, install, and report the results of `fgrep serial /var/log/stratux.log`

peepsnet:
Results from execuiting commands

pi@raspberrypi ~ $ sudo su -
root@raspberrypi:~# ps auxww | grep dump1090
root      2201  0.0  0.1   3560  1796 pts/0    S+   20:47   0:00 grep dump1090
root@raspberrypi:~#

I will ado the rest and report

peepsnet:
And here is the rest

pi@raspberrypi ~ $ sudo su -
root@raspberrypi:~# fgrep serial /var/log/stratux.log
2016/02/21 20:54:30 ===== UAT Device serial: stx:0:-28 =====
root@raspberrypi:~#

peepsnet:
The per sdr ppm logic is incorrect in some manor. If I set my serials to stx:978:-24 and stx:1090:-28 then the system boots properly and I get both sdr's receivng data.

peepsnet:
[...]

PR #278, opened 2016-02-22
Enhancements to dual-PPM (minor bug fix, web status)

yihchun:
- Skipping the stratux serial check when the other device is already bound. This prevents a misconfiguration (two SDRs with stratux:1090, for example) from blocking the second frequency from coming online.
- Fix a special-case bug where id 0 is flexible and id 1 is hard-wired to UAT. Comments explain why the same special-case need not be applied to ES hard-wiring.
- Added logging of serial numbers and PPMs per my comment on issue #275
- Added SDRs active count to status page

### DIFF
--- a/main/sdr.go
+++ b/main/sdr.go
@@ -404,6 +404,23 @@ func sdrWatcher() {
 						id = 1
 					}
 				}
+
+				/* Address #275. If both devices are nil, but
+			 	 * id 1 matches the rUAT string, we need to
+				 * force the UAT to id 1, so that UAT doesn't
+				 * consume a flexible id 0 and leave ES bound
+				 * to an undesired, fixed 978.
+				 * A parallel fix is not installed in the ES
+				 * code, because both cases are already handled:
+				 * id 0 flex, id 1 1090, then 0 bound to UAT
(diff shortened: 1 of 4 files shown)


jpoirier:
fyi - you'll need resubmit the appropriate parts once I get request #279 submitted; there are some significant changes

main/sdr.go
```diff
@@ -147,12 +149,14 @@ func getPPM(serial string) int {
 
 func (e *ES) sdrConfig() (err error) {
```
peepsnet: Why is this function so empty compared to the func (u *UAT) sdrConfig() (err error)

jpoirier: we created our own Go wrapper around the dump978 tool (and compile it as a library) so we're able to configure the dongle internally (using gortlsdr) whereas we call the external dump1090 binary via exec.Command (shell command) and pass it options which it uses to configure the dongle.

jpoirier:
@yihchun        it'd be great if you could review #279

main/sdr.go
peepsnet: where is dump1090 called??

yihchun:
Closing; will reimplement after #279 is merged.

Issue #239, opened 2016-02-06
prioritizing a dongle with a valid id string over an anonymous one 


jpoirier:
when a dongle with a stratux id string is introduced, or exists in the config list, it should override and/or take the place of an anonymous dongle and the anonymous dongle should then be configured for an unused frequency.

edit: behavior of unused anonymous dongle

bkwny:
And if the newly-freed anonymous dongle could be used to satisfy an frequency enabled on the WebUI but not having a dongle, put the anonymous dongle to use for it.

jpoirier:
On Mon, Feb 8, 2016 at 8:27 PM, bkwny <notifications@github.com> wrote:

> And if the newly-freed anonymous dongle could be used to satisfy an
> frequency enabled on the WebUI but not having a dongle, put the anonymous
> dongle to use for it.
>
thanks.

I updated the initial post's verbiage. Though the current code will
re-config the anonymous dongle either in the current pass or the next pass,
depending on what frequency the dongle with the id was configured for, it's
not a bad idea to explicitly state the requirement .

PR #279, opened 2016-02-22
better anonymous & tagged dongle handling, sdr file cleanup

jpoirier:
This fixes/simplifies the handling of anonymous and tagged dongles.

### DIFF
--- a/main/sdr.go
+++ b/main/sdr.go
@@ -405,7 +405,7 @@ func sdrWatcher() {
 					serial = ""
 				}
 
-				if !!rES.hasID(serial) {
+				if !rES.hasID(serial) {
 					UATDev = &Device{indexID: id, serial: serial, wg: &sync.WaitGroup{}}
 					if err := UATDev.sdrConfig(); err != nil {
 						log.Printf("UATDev = &UAT{indexID: id} failed: %s\n", err)
@@ -447,7 +447,7 @@ func sdrWatcher() {
 					serial = ""
 				}
 
-				if !!rUAT.hasID(serial) {
+				if !rUAT.hasID(serial) {
 					ESDev = &Device{indexID: id, serial: serial, wg: &sync.WaitGroup{}}
 					if err := ESDev.sdrConfig(); err != nil {
 						log.Printf("ESDev = &ES{indexID: id} failed: %s\n", err)
--- a/main/sdr.go
+++ b/main/sdr.go
@@ -31,7 +31,7 @@ type Device struct {
 	indexID int
 	ppm     int
 	serial  string
-	hasID   bool
+	idSet   bool
 }
 
 type UAT Device
@@ -240,6 +240,7 @@ func (u *UAT) sdrConfig() (err error) {
 	//---------- Get/Set Freq Correction ----------
 	freqCorr := u.dev.GetFreqCorrection()
 	log.Printf("\tGetFreqCorrection: %d\n", freqCorr)
+
 	u.ppm = getPPM(u.serial)
 	err = u.dev.SetFreqCorrection(u.ppm)
 	if err != nil {
@@ -318,8 +319,10 @@ func reCompile(s string) *regexp.Regexp {
 }
 
 type regexUAT regexp.Regexp
+type regexES regexp.Regexp
 
 var rUAT = (*regexUAT)(reCompile("str?a?t?u?x:978"))
+var rES = (*regexUAT)(reCompile("str?a?t?u?x:1090"))
 
 func (r *rUAT) hasID(serial string) bool {
 	if r == nil {
@@ -328,10 +331,6 @@ func (r *rUAT) hasID(serial string) bool {
 	return (*regexp.Regexp)(r).MatchString(serial)
 }
 
-type regexES regexp.Regexp
-
-var rES = (*regexUAT)(reCompile("str?a?t?u?x:1090"))
-
 func (r *rES) hasID(serial string) bool {
 	if r == nil {
 		return strings.HasPrefix(serial, "stratux:1090")
@@ -339,8 +338,71 @@ func (r *rES) hasID(serial string) bool {
 	return (*regexp.Regexp)(r).MatchString(serial)
 }
 
+func createUATDev(id int, serial string, idSet bool) error {
+	UATDev = &UAT{indexID: id, serial: serial}
+	if err := UATDev.sdrConfig(); err != nil {
+		log.Printf("UATDev.sdrConfig() failed: %s\n", err)
+		UATDev = nil
+		return err
+	}
+	UATDev.wg = &sync.WaitGroup{}
+	UATDev.idSet = idSet
+	UATDev.closeCh = make(chan int)
+	UATDev.wg.Add(1)
+	go UATDev.read()
+	return nil
+}
+
+func createESDev(id int, serial string, idSet bool) error {
+	ESDev = &ES{indexID: id, serial: serial}
+	if err := ESDev.sdrConfig(); err != nil {
+		log.Printf("ESDev.sdrConfig() failed: %s\n", err)
+		ESDev = nil
+		return err
+	}
+	ESDev.wg = &sync.WaitGroup{}
+	ESDev.idSet = idSet
+	ESDev.closeCh = make(chan int)
+	ESDev.wg.Add(1)
+	go ESDev.read()
+	return nil
+}
+
+func configDevices(count int, es_enabled, uat_enabled bool) {
+	// entry to this function is only valid when both UATDev and ESDev are nil
+	for i := 0; i < count; i++ {
+		_, _, s, err := rtl.GetDeviceUsbStrings(i)
+		if err == nil {
+			if uat_enabled && UATDev == nil && rUAT.hasID(s) {
+				createUATDev(i, s, true)
+				continue
+			}
+			if es_enabled && ESDev == nil && rES.hasID(s) {
+				createESDev(i, s, true)
+				continue
+			}
+
+			// we only get here when it's an anonymous dongle
+			if uat_enabled && UATDev == nil {
+				createUATDev(i, s, false)
+				continue
+			}
+			if es_enabled && ESDev == nil {
+				createESDev(i, s, false)
+				continue
+			}
+		} else {
+			log.Printf("rtl.GetDeviceUsbStrings id %d: %s\n", i, err)
+		}
+	}
+}
+
 // Watch for config/device changes.
 func sdrWatcher() {
+	prevCount := 0
+	prevUAT_Enabled := false
+	prevES_Enabled := false
+
 	for {
 		time.Sleep(1 * time.Second)
 		if sdrShutdown {
@@ -354,6 +416,7 @@ func sdrWatcher() {
 			}
 			return
 		}
+
 		count := rtl.GetDeviceCount()
 		atomic.StoreUint32(&globalStatus.Devices, uint32(count))
 
@@ -362,7 +425,7 @@ func sdrWatcher() {
 			count = 2
 		}
 
-		// cleanup if necessary
+		// check for either no dongles or none enabled
 		if count < 1 || (!globalSettings.UAT_Enabled && !globalSettings.ES_Enabled) {
 			if UATDev != nil {
 				UATDev.shutdown()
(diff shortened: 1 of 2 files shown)


peepsnet:
There was talk on prioritizing SDR choices in the case of an SDR failure. 

Assume 2 freq tuned antennas and dual SDR's
If an SDR fails in flight I believe the logical choice is to assign the working SDR to the 978 freq(if turned on in settings) so weather/nexrad is still available(1090 is shutdown and 978 assigned).  

Does this code keep the stx:978:-24 and stx:1090:-28 or stx:0:21per SDR PPM serial format??

The 0 in the serial simply means I do not care which freq you assign to it but I want the ppm setting

If my SDR's serials were set like this:
stx:978:-24 and stx:1090:-28
or
stx:978:-24 and stx:0:24

I want each SDR to be set to the freq in the serial or if a 0 then set other freq first then what is left is assigned to the 0 serial SDR

IF I have 978 turned on in the settings and no 978 serial is found then set the remaining/only SDR to 978 regardless of the serials

Does this make sense??

jpoirier:
> Does this code keep the stx:978:-24 and stx:1090:-28 or stx:0:21per SDR PPM serial format??

yes

> stx:978:-24 and stx:1090:-28

these should both get identified and configured as tagged dongles 

> stx:978:-24 and stx:0:24

stx:978:-24 should get tagged and configured as a 978 dongle and stx:0:24 should get tagged as an anonymous dongle and configured to 1090

> I want each SDR to be set to the freq in the serial or if a 0 then set other freq first then what is left is assigned to the 0 serial SDR

yes, this code should do that 

> IF I have 978 turned on in the settings and no 978 serial is found then set the remaining/only SDR to 978 regardless of the serials

this code won't do that; it prioritizes the serial id regardless of the dongle count

peepsnet:
so If I have a stx:978:24 dongle and a stx:0:-21 dongle and the 978 dongle fails... that happens on reboot??

jpoirier:
> What if I have a stx:978:55 dongle and only the 1090 SDR turned on in the settings what happens on boot??

if *only* 1090 is turned on in the web ui settings and there's a single dongle with the id "stx:978:55" it won't be configured  

> so If I have a stx:978:24 dongle and a stx:0:-21 dongle and the 978 dongle fails... that happens on reboot??

assuming the stx:978:24 dongle is dead, ie it doesn't show up in the device count, the stx:0:-21 dongle will be configured to 978 provided 978 is enabled in the web ui settings and *regardless* of whether 1090 is enabled

peepsnet:
great. That is what I wanted to know.

peepsnet:
Great. Thats what I wanted to know!

main/sdr.go
```diff
@@ -319,20 +312,96 @@ func sdrKill() {
...
+				createUATDev(i, s, true)
+				continue
+			}
+			if es_enabled && ESDev == nil && rES.hasID(s) {
+				createESDev(i, s, true)
+				continue
+			}
+
+			// we only get here when it's an anonymous dongle
+			if uat_enabled && UATDev == nil {
```
yihchun: This still has the bug that I fixed in #278: if i 0 is anonymous and i 1 is wired to UAT, then only 0 will be activated on UAT.

The fix is to run the loop twice: first to assign non-anonymous dongles and the second to assign anonymous dongles. Ideally a third loop would assign all dongles, in case a user erroneously assigned both dongles to :1090. This logic would also implement #239.

main/sdr.go
```diff
@@ -319,20 +312,96 @@ func sdrKill() {
...
+	// note , compile returns a nil pointer on error
+	r, _ := regexp.Compile(s)
+	return r
+}
+
+type regexUAT regexp.Regexp
+type regexES regexp.Regexp
+
+var rUAT = (*regexUAT)(reCompile("str?a?t?u?x:978"))
+var rES = (*regexUAT)(reCompile("str?a?t?u?x:1090"))
```
yihchun: Do you mean (*regexES)?

main/sdr.go
```diff
@@ -244,9 +236,12 @@ func (u *UAT) sdrConfig() (err error) {
 	} else {
 		log.Printf("\tResetBuffer Successful\n")
 	}
+
 	//---------- Get/Set Freq Correction ----------
 	freqCorr := u.dev.GetFreqCorrection()
 	log.Printf("\tGetFreqCorrection: %d\n", freqCorr)
+
+	u.ppm = getPPM(u.serial)
```
yihchun: Suggest to add 
`log.Printf("===== UAT Device Serial: %s PPM %d =====\n", u.serial, u.ppm)`

main/sdr.go
```diff
@@ -150,9 +139,17 @@ func (e *ES) sdrConfig() (err error) {
 	return
```
yihchun: Suggest to add `log.Printf("===== ES Device Serial: %s PPM %d =====\n", e.serial, e.ppm)`

main/sdr.go
peepsnet: I love the copy and paste errors!!  They always kill me!

main/sdr.go
jpoirier: added a print for serial at the top and ppm was already printed out for freq correction

main/sdr.go
jpoirier: done

main/sdr.go
jpoirier: You're right, and thanks for the review! I'll fix and push it later today.

jpoirier:
@yihchun  it should be be working now but let me know if you see any (more) problems/flaws

main/sdr.go
```diff
@@ -319,21 +314,164 @@ func sdrKill() {
...
+			if uat_enabled && UATDev == nil {
+				createUATDev(v, s, false)
+			} else if es_enabled && ESDev == nil {
+				createESDev(v, s, false)
+			}
+		} else {
+			log.Printf("rtl.GetDeviceUsbStrings id %d: %s\n", v, err)
+		}
+	}
+}
```
yihchun: Looks great but I think your big comment block is incorrect; since loop 2 doesn't check, for example,
`if uat_enabled && UATDev == nil && !rES.hasID(s)`
loop 2 will automatically override (ignore?) any dual-assignments. In my opinion, ideal behavior would include such a check in loop 2 and then override that check in a loop 3, but the existing code is very close to ideal behavior. A case in which this might arise is if someone connected 4 SDRs: one 1090 connected to a ES antenna, two 978s connected to UAT antennas, and one anonymous connected to a generically-lengthed antenna, and the 1090 SDR fails; the existing code might assign the second 978 to ES when the correct choice may be the anonymous one.

But the code is already ideal for my setup.

jpoirier: On Mon, Feb 22, 2016 at 11:39 AM, yihchun <notifications@github.com> wrote:

> In main/sdr.go
> <https://github.com/cyoung/stratux/pull/279#discussion_r53661410>:
>
> >  	}
> >
> > +	// loop 2; assign anonymous dongles
> > +	for _, v := range unusedIDs {
> > +		_, _, s, err := rtl.GetDeviceUsbStrings(v)
> > +		if err == nil {
> > +			if uat_enabled && UATDev == nil {
> > +				createUATDev(v, s, false)
> > +			} else if es_enabled && ESDev == nil {
> > +				createESDev(v, s, false)
> > +			}
> > +		} else {
> > +			log.Printf("rtl.GetDeviceUsbStrings id %d: %s\n", v, err)
> > +		}
> > +	}
> > +}
>
>
one caveat is that we only support up to two dongles, the first two seen by
librtlsdr (device id 0 and 1) :)

        // support two and only two dongles
        if count > 2 {
            count = 2
        }

> Looks great but I think your big comment block is incorrect; since loop 2
> doesn't check, for example,
> if uat_enabled && UATDev == nil && !rES.hasID(s)
> loop 2 will automatically override (ignore?) any dual-assignments. In my
> opinion, ideal behavior would include such a check in loop 2 and then
> override
>
I'm not sure I understand.

Once loop 1 finishes the only remaining *enabled* dongles *must* be
anonymous, therefore, in loop 2, rES.hasID(s) and rUAT.hasID(s) could never
be true for an enabled and as yet unassigned device.

unusedIDs contains no IDs prior to loop 1 and device ID 0 is only appended
to unusedIDs when said dongle has neither a 1090 or 978 serial, similar for
device ID 1. Therefore, when loop 2 starts we only range over device IDs
that weren't consumed in loop 1.

main/sdr.go
jpoirier: On Mon, Feb 22, 2016 at 7:39 AM, peepsnet <notifications@github.com> wrote:

> In main/sdr.go
> <https://github.com/cyoung/stratux/pull/279#discussion_r53624975>:
>
> > +type regexUAT regexp.Regexp
> > +type regexES regexp.Regexp
> > +
> > +var rUAT = (*regexUAT)(reCompile("str?a?t?u?x:978"))
> > +var rES = (*regexUAT)(reCompile("str?a?t?u?x:1090"))
>
> I love the copy and paste errors!! They always kill me!
>
I'm usually pretty good about following a fairly strict workflow to avoid
this sort of stuff... :(

main/sdr.go
yihchun: > one caveat is that we only support up to two dongles 

OK, I forgot about that force `count = 2`. I thought you were going to fix #239 also. At this point I don't see any significant reason to do that force.

>  Once loop 1 finishes the only remaining *enabled* dongles *must* be
anonymous

I'm not sure I see it that way. If both id 0 and id 1 have rES.hasID(s), then in round 1, id 0 will be wired to ES, then the next iteration id 1 will not satisfy `ESDev == nil` so it will fall through to the else, thus being added to the `unusedIDs` array. At this point you'll send that one to UAT, which is correct behavior but id 1 is certainly not anonymous.

jpoirier: On Mon, Feb 22, 2016 at 12:31 PM, yihchun <notifications@github.com> wrote:

> In main/sdr.go
> <https://github.com/cyoung/stratux/pull/279#discussion_r53668876>:
>
> >  	}
> >
> > +	// loop 2; assign anonymous dongles
> > +	for _, v := range unusedIDs {
> > +		_, _, s, err := rtl.GetDeviceUsbStrings(v)
> > +		if err == nil {
> > +			if uat_enabled && UATDev == nil {
> > +				createUATDev(v, s, false)
> > +			} else if es_enabled && ESDev == nil {
> > +				createESDev(v, s, false)
> > +			}
> > +		} else {
> > +			log.Printf("rtl.GetDeviceUsbStrings id %d: %s\n", v, err)
> > +		}
> > +	}
> > +}
>
> one caveat is that we only support up to two dongles
>
> OK, I forgot about that force count = 2. I thought you were going to fix
> #239 <https://github.com/cyoung/stratux/issues/239> also. At this point I
> don't see any significant reason to do that force.
>
> Once loop 1 finishes the only remaining *enabled* dongles *must* be
> anonymous
>
> I'm not sure I see it that way. If both id 0 and id 1 have rES.hasID(s),
> then in round 1, id 0 will be wired to ES, then the next iteration id 1
> will not satisfy ESDev == nil so it will fall through to the else, thus
> being added to the unusedIDs array. At this point you'll send that one to
> UAT, which is correct behavior but id 1 is certainly not anonymous.
>

Lol. Okay, it only took you what 4 or 5 times to explain the dual
assignment scenario...thanks! I'll add the additional checks. :)

peepsnet: I like the term freq-flexible. I hope to contribute to the faq when this per SDR PPM is working.

peepsnet:
@jpoirier How can I pull these changes to my stratux. just your changed files and test

jpoirier:
On Mon, Feb 22, 2016 at 12:55 PM, peepsnet <notifications@github.com>
wrote:

> @jpoirier <https://github.com/jpoirier> How can I pull these changes to
> my stratux. just your changed files and test
>
I think this should work

Assuming you have the latest stratux code, you can create and checkout a
new branch then: git pull origin pull/279/head

It'll merge everything then you can make, make install...

peepsnet:
Ok it appears merged based on the new log entries. but it assigned the SDRs backwards

jpoirier:
On Mon, Feb 22, 2016 at 1:24 PM, peepsnet <notifications@github.com> wrote:

> Ok it appears merged based on the new log entries. but it assigned the
> SDRs backwards
>
The pull request (https://github.com/cyoung/stratux/pull/280) is still open
so unless you explicitly pull and merge it yourself you won't have the
modified code.

What do you mean assigned backwards?

peepsnet:
It is working correctly... I didn't reboot the PI after changing the Serials. I just restarted the stratux service.

jpoirier:
On Mon, Feb 22, 2016 at 1:50 PM, peepsnet <notifications@github.com> wrote:

> It is working correctly... I didn't reboot the PI after changing the
> Serials. I just restarted the stratux service.
>
Awesome, and thanks for taking the time to test this out!

yihchun:
Looks great!

PR #280, opened 2016-02-22
Change to be more specific in choosing update file

peepsnet:
This change might prevent another update*.sh script from being executed since it is just looking in the root directory for the latest file with the word update. 

This would require naming the update the same update\*stratux\*v*.sh
i.e.: update-stratux-v0.7b4-f98a8d58c6.sh

Just a bit more safety...


cyoung:
Good change, thank you.

peepsnet:
Just so excited that I have actually contributed to the project!!!

PR #281, opened 2016-02-22
Better page reloading

peepsnet:
This should help for a better page reload. 

Adding a random number to the URL so the browser things its a new page when the STRATUX link is clicked

### DIFF
--- a/web/index.html
+++ b/web/index.html
@@ -106,7 +106,7 @@
 		<!-- Create Header Navbar with left Menu Button -->
 		<div class="navbar navbar-app navbar-absolute-top">
 			<div class="navbar-brand navbar-brand-center" ui-yield-to="title">
-				<a href="#/">
+				<a href="/?1" onClick="this.href=this.href.split('?')[0]+'?'+new Date().getTime()">
                     Stratux
             </a>
 			</div>


bradanlane:
won't this cause unnecessary fetching of resources by the browser? 

I saw an error my my desktop browser (Chrome) when it attempted to reference the manifest. I'm looking into what/why. Solving the manifest cache will be a more efficient mechanism for invalidating old resources.

peepsnet:
Well that might be but there is an issue with some browsers and the #.  It wants to just jump to an anchor and not actually fetch the page again

bradanlane:
what browser(s) are you having issues with?

peepsnet:
I am using chrome/firefox but I have had issues with the connected green bar coming back after just clicking on the link. If i refresh the page I get it back. 

This is more a way for a dev to make sure they get a new page while working with stratux.  The load on the device is minimal.

peepsnet:
I guess my issue is when I am on the status page and reboot my stratux from SSH I just wanted a way to force a good reload of the data. The # in the url I believe is the issue. Also a cached page might not want to reload all data because the url is the same.

RobertGary1:
Usually you'd want to limit cache busting to a version number so you can be sure that in the majority of field cases (when the version hasn't changed) you'll get the performance benefit of the browser cache.

bradanlane:
we can move `index.html` out of `appcache` by placing under a section labeled `NETWORK` in teh manifest. alternately, we can just disable all cross session caching if its not a performance hit.

cyoung:
@peepsnet: could you make a screen recording of the issue that this solves? I don't understand what the issue is that this addresses.

peepsnet:
If I can... It was just the /#/ in the url would sometimes not reload the page as if it was jumping to an anchor and not actually loading a page. If i see it stuck again I will try to do it.

peepsnet:
Well I guess it was a fix for an old issue. I am not able to get this to work cross-browser

PR #285, opened 2016-02-24
Dump1090 improvements. Overhauled traffic parsing. UAT groundspeed and air/ground status bug fixes.

AvSquirrel:
Add Stratux-specific branch of MalcolmRobb/dump1090 and overhaul the traffic.go parsing routines.

**Dump1090 changes:**
- Add parsing of 1090ES TIS-B messages
- Add parsing of ADS-B / ADS-R / TIS-B NACp category from DF=17, TC=29/31 and DF=18, TC=19 messages
- Add parsing of GNSS altitude flag
- Add parsing of delta height between GNSS altitude and baro altitude
- New JSON-formatted output of fields used by traffic.go are sent on TCP port 30006. These include additions for NACp, signal strength, delta height, and GNSS height flag as noted in the [dump1090 README](https://github.com/AvSquirrel/dump1090#port-30006).

**Stratux changes:**
1090 ES parsing is completely overhauled to support the new dump1090 format. Additional ES parsing changes include:
- dump1090 raw signal strength is reported
- NIC is now calculated from DF17 position message type code
- NACp is reported when available (else, it is set to NIC, because not all emitters send this characteristic)
- Addr_type (UAT address qualifier) is set per the presence of DF=17 or DF=18 with the appropriate Control Field flag. This will allow GDL90-connected devices to identify whether targets are ADS-B, ADS-R, or TIS-B
-New TargetType flag will allow traffic source to be cleanly shown on UI traffic page (future commit)

978 UAT parsing improved to fix a couple of significant bugs:
- ti.OnGround initializes to 'false' when new ti is created, but is never explicitly set to false when vaild airborne messages are received. If a 978 target is ever seen on the ground, it will never be set to flight status.
- ew_vel and ns_vel changed from int16 to int32 to solve the zero airspeed issue noted in #277

Other 978 improvements include:
- NIC and NACp set from UAT message
- UAT signal strength is now passed to the traffic decoder and is tracked as part of each traffic target (similar to changes for 1090ES)

Refinements to GDL90 message packaging include:
- Track type set to "true track" rather than "true heading" (latter applies mainly to surface vehicles). 
- Add support for setting "Extrapolated position" flag for future traffic coasting algorithm
- Report actual NIC / NACp of targets, rather than hardcoded "1"

Other changes include:
- Traffic database is printed to stratux.log every ten seconds as a debug / troubleshooting aid. This frequency may be reduced or turned into a selectable debug option in future releases.
[...]

Issue #171, opened 2015-12-29
Stratux displaying altitude of certain aircaft as "zero"


Ergonomicmike:
I'll make it official, based on Scrabo Tower's test with a SkyRadar and a Stratux in the Phoenix area:
"Yeah it's Stratux , N971PA just passed by, SkyRadar showed altitude, Stratux was zero."

https://www.reddit.com/r/stratux/comments/3yld3s/video_of_flight_school_ac_not_displaying_altitude/cyfcux3

short video of the issue at https://vid.me/gzLT

cyoung:
I looked at your 0002-es.log.gz and there's only messages from UAL1640. Have you got replay logs for N971PA?

Axtel4:
Here is the log I downloaded from his Wiki link.  It has two of the "PA" aircraft; N926PA (Mode S = ACD5BA) and N925PA (Mode S = ACD203). There are also 3 other aircraft that report an altitude of 0 ft.


[0002-es.zip](https://github.com/cyoung/stratux/files/74394/0002-es.zip)

cyoung:
```
$ cat 0002-es.log | grep MSG,3 | cut -d, -f2-19  | egrep -v '(ACD5BA|ACD203)'
MSG,3,,,A73DB6,,,,,,,35000,,,,,,
MSG,3,,,A73DB6,,,,,,,35000,,,,,,
MSG,3,,,A73DB6,,,,,,,35000,,,33.30645,-111.80327,,
MSG,3,,,A73DB6,,,,,,,35025,,,33.31256,-111.77715,,
MSG,3,,,A73DB6,,,,,,,35025,,,33.31256,-111.77715,,
MSG,3,,,A73DB6,,,,,,,35025,,,33.31256,-111.77715,,
MSG,3,,,A73DB6,,,,,,,35025,,,33.31256,-111.77715,,
MSG,3,,,A73DB6,,,,,,,35000,,,33.32153,-111.73970,,
MSG,3,,,A73DB6,,,,,,,35000,,,33.32185,-111.73700,,
MSG,3,,,A73DB6,,,,,,,35025,,,33.32222,-111.73689,,
MSG,3,,,A73DB6,,,,,,,35025,,,33.32293,-111.73124,,
MSG,3,,,A73DB6,,,,,,,35025,,,33.32350,-111.73123,,
MSG,3,,,A73DB6,,,,,,,35025,,,33.32386,-111.72815,,
[...]

PR #292, opened 2016-02-27
Incorporate dump1090-mutability; improved signal strength reporting; bug fixes

AvSquirrel:
Improvements:

- Incorporate dump1090 changes from #285 into a recent branch of dump1090-mutability. This is an active branch of dump1090, and has added additional functionality over the MalcolmRobb branch,
- Traffic UI now shows if ground-based traffic transmissions are TIS-B or ADS-R on both UAT and 1090ES.
- Improvements to signal strength reporting. All information sources (1090ES traffic, UAT traffic, and UAT FIS-B uplinks) are now displayed with received power converted to dB and scaled the same. Minimum signal detection occurs at about -40 dB, with various messages detectable at higher powers. Short Mode S messages (e.g. interrogation replies) are the easiest to detect, and can be seen starting at about -35 dB, and ADS-B / FIS-B type messages will be seen starting at -30 dB, and reliably above -25 dB. Nominal saturation occurs at 0.0 dB, and the strongest signals will truncate at about +1.4 dB.
- Create and check for dump1090 heartbeat message
- Towers page refreshes more quickly, and shows current, average, and peak power (dB) levels. (Old traffic page shows strength as sum of amplitude.)
- Allow UAT emitters with NIC = 0 into the traffic data stream. This will make uncertified and/or misconfigured UAT transmitters visible.
- Assorted bug fixes

To-do:
- Validate performance of new 1090ES TIS-B support
- Add CSV-out option (or converter tool) for translating JSON-formatted traffic into a format easier for @Axtel4 to parse
- "Signal bar" mappings for dB signal strength. 
- Merge dump1090 improvements into mutability's master branch

image

image

### DIFF
--- a/main/traffic.go
+++ b/main/traffic.go
@@ -152,7 +152,7 @@ func sendTrafficUpdates() {
 	defer trafficMutex.Unlock()
 	cleanupOldEntries()
 	var msg []byte
-	if (stratuxClock.Time.Second() % 10) == 0 {
+	if (globalSettings.DEBUG && stratuxClock.Time.Second()%10) == 0 {
 		log.Printf("List of all aircraft being tracked:\n")
 		log.Printf("==================================================================\n")
 	}
--- a/main/traffic.go
+++ b/main/traffic.go
@@ -561,7 +561,12 @@ func esListen() {
 			var newTi *dump1090Data
 			err = json.Unmarshal([]byte(buf), &newTi)
 			if err != nil {
-				log.Printf("can't read ES traffic information: %s\n", err.Error())
+				log.Printf("can't read ES traffic information from %s: %s\n", buf, err.Error())
+				continue
+			}
+
+			if (newTi.Icao_addr & 0xFF000000) != 0 { //24-bit overflow is used to signal heartbeat
+				log.Printf("No traffic last 60 seconds. Heartbeat message from dump1090: %s\n", buf)
 				continue
 			}
 
--- a/main/traffic.go
+++ b/main/traffic.go
@@ -841,19 +841,22 @@ func updateDemoTraffic(icao uint32, tail string, relAlt float32, gs float64, off
 
 	ti.Icao_addr = icao
 	ti.OnGround = false
-	ti.Addr_type = uint8(icao % 4) // 0 == ADS-B; 1 == reserved; 2 == TIS-B with ICAO address; 3 == TIS-B without ICAO address.
-	if ti.Addr_type == 1 {         //reserved value
-		ti.Addr_type = 0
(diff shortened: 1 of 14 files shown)


main/traffic.go
```diff
@@ -835,6 +845,17 @@ func updateDemoTraffic(icao uint32, tail string, relAlt float32, gs float64, off
 	if ti.Addr_type == 1 {         //reserved value
 		ti.Addr_type = 0
 	}
+
+	if ti.Addr_type == 0 {
+		ti.TargetType = TARGET_TYPE_ADSB
+	}
+	if ti.Addr_type == 2 {
```
jpoirier: these two checks can be done with an "if else" check 

I guess if ti.Addr_type isn't 0 or 2 or 3 then the default for ti.TargetType is TARGET_TYPE_MODE_S (TARGET_TYPE_MODE_S is 0 which is what ti.TargetType is initialized to). It'd be clearer if the last check was an else that explicitly defaults to the desired value. E.g.

        if ti.Addr_type == 0 {
		ti.TargetType = TARGET_TYPE_ADSB
	} else if ti.Addr_type == 2 {
		ti.TargetType = TARGET_TYPE_ADSR
	} else if ti.Addr_type == 3 {
		ti.TargetType = TARGET_TYPE_TISB
	} else {
                ti.TargetType = TARGET_TYPE_MODE_S // or should this be something else
        }

AvSquirrel:
Flew today, and realized I'm not handling the CA field in dump1090-mutability quite right for DF18 traffic. Stand by for a new dump1090 submodule update.

@jpoirier -- noted. Though it is a function I'm using solely to generate "demo" targets (and therfore the default case isn't too critical) it's worth cleaning up.

main/traffic.go
AvSquirrel: Addressed with 769169d; copied logic from `parseDownlinkReport()`

AvSquirrel:
CA/CF fix included on commit 769169d. Validated against the airborne message types (DF0, DF4, DF5, DF11, DF17); may need to flight test with a 1090-In airplane to validate DF18.

cyoung:
@AvSquirrel - Q-bit handling in this branch? Have not reviewed the mutability branch at all, let's give it a try anyway. Did you enable oversampling?

https://github.com/cyoung/stratux/issues/171#issuecomment-167903450

AvSquirrel:
Ayup.

Oversampling is enabled in the exec call in sdr.go, and q-bit is supported. (The mutability branch is based on the last the most MalcolmRobb commit, and is now 497 commits ahead of it.)

cyoung:
Aha. Sounds like a winner.

PR #297, opened 2016-02-29
fix missing paren

jpoirier:
This fixes a fix that I unfixed


cyoung:
Already re-fixed it!

jpoirier:
hmm, thought I'd messed up the fix...too much happening in parallel :)

cyoung:
You did, and the build failed and I noticed it ... so I committed the same change twice.

jpoirier:
k. thanks

PR #319, opened 2016-03-10
spindle: proper lib prefix

jpoirier:


jpoirier:
Some platforms know to install fftw3 as libfftw3 but others require the unadorned fftw3. Go figure.

PR #321, opened 2016-03-11
sdr: capture global state info immediately after it's checked

jpoirier:


jpoirier:
closing in lieu of broader global var changes

PR #322, opened 2016-03-11
proper read/write synchronization for globals

jpoirier:
should probably be similar for all globals, if you're happy with the PR we can do the same for the remaining ones


jpoirier:
This isn't going to work on a newer Go compiler unless the defaultSettings/newSettings is reworked.

PR #338, opened 2016-03-16
auto init on package import

jpoirier:


cyoung:
If we end up breaking up different parts of "package main" into sub-packages, we have to take care that there is no race condition between `init()` calls. There doesn't seem to be a way to control the order that they are called in.

jpoirier:
Items in the package init function should be strictly package specific, e.g. create a lut for a crc, etc..., so order shouldn't matter, and everything else should be in an exported Init function.

PR #344, opened 2016-03-22
Tweak error messages to use FontAwesome icon and stock angular maui colors.

egid:
Also added in a `className` so the messages can be explicitly styled later, although that may not be necessary.


egid:
Here's a preview of how it looks, albeit in a local test environment and not running straight off a stratux:

cyoung:
It's a beauty!

PR #365, opened 2016-04-03
Add missing fields for multiple-row INSERT

AvSquirrel:
The SQLite `INSERT` in `bulkInsert()` was missing fields for all but the first row of each insert. This was what caused the log file to only contain one data point each ten seconds for the queueable tables.

To fix this, I add a  missing string concatenate to `stmt` to allow additional rows to be inserted.

I also had to limit the number of rows per insert to 30. This keeps the total number of columns under 1000, which seems to be the limit for this version of SQLite. To keep the queue clear, write ticker was reduced from 10 sec to 0.5 sec.

### DIFF
--- a/main/datalog.go
+++ b/main/datalog.go
@@ -216,16 +216,17 @@ func bulkInsert(tbl string, db *sql.DB) (res sql.Result, err error) {
 
 	batchVals := insertBatchIfs[tbl]
 	for len(batchVals) > 0 {
-		i := int(0) // Maximum of 25 rows per INSERT statement.
+		i := int(0)
 		stmt := ""
 		vals := make([]interface{}, 0)
-		querySize := uint64(0) // Size of the query in bytes.
-		for len(batchVals) > 0 && i < 1000 && querySize < 750000 {
+		querySize := uint64(0)                                   // Size of the query in bytes.
+		for len(batchVals) > 0 && i < 30 && querySize < 750000 { // Maximum of 1000? columns for INSERT on default sqlite compile. trafficData is limiting function with 31 parameters.
 			if len(stmt) == 0 { // The first set will be covered by insertString.
 				stmt = insertString[tbl]
 				querySize += uint64(len(insertString[tbl]))
 			} else {
 				addStr := ", (" + strings.Join(strings.Split(strings.Repeat("?", len(batchVals[0])), ""), ",") + ")"
+				stmt += addStr
 				querySize += uint64(len(addStr))
 			}
 			for _, val := range batchVals[0] {
@@ -235,9 +236,10 @@ func bulkInsert(tbl string, db *sql.DB) (res sql.Result, err error) {
 			batchVals = batchVals[1:]
 			i++
 		}
-		//		log.Printf("inserting %d rows to %s. querySize=%d\n", i, tbl, querySize)
+		// log.Printf("inserting %d rows to %s. querySize=%d\n", i, tbl, querySize) // debug
 		res, err = db.Exec(stmt, vals...)
 		if err != nil {
+			log.Printf("sqlite insert error: '%s'\n", err.Error())
 			return
 		}
 	}
@@ -338,7 +340,7 @@ func dataLogWriter(db *sql.DB) {
 	dataLogWriteChan = make(chan DataLogRow, 10240)
 	// The write queue. As data comes in via dataLogChan, it is timestamped and stored.
 	//  When writeTicker comes up, the queue is emptied.
-	writeTicker := time.NewTicker(10 * time.Second)
+	writeTicker := time.NewTicker(500 * time.Millisecond)
 	rowsQueuedForWrite := make([]DataLogRow, 0)
 	for {
 		select {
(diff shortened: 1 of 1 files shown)


main/datalog.go
```diff
@@ -216,16 +216,17 @@ func bulkInsert(tbl string, db *sql.DB) (res sql.Result, err error) {
...
 	batchVals := insertBatchIfs[tbl]
 	for len(batchVals) > 0 {
-		i := int(0) // Maximum of 25 rows per INSERT statement.
+		i := int(0)
 		stmt := ""
 		vals := make([]interface{}, 0)
-		querySize := uint64(0) // Size of the query in bytes.
-		for len(batchVals) > 0 && i < 1000 && querySize < 750000 {
+		querySize := uint64(0)                                   // Size of the query in bytes.
+		for len(batchVals) > 0 && i < 30 && querySize < 750000 { // Maximum of 1000? columns for INSERT on default sqlite compile. trafficData is limiting function with 31 parameters.
```
AvSquirrel: @cyoung -- I'll say up front that this isn't going to be a large enough query size for `dump1090_terminal` or `traffic` if you're tracking a decent number of targets, even at 0.5 second updates... but we literally can not set this larger than 30 or 31 with our current traffic struct due to the 1000 column query limit.

What do you think about rolling the `traffic` updates back to once per second, and checking the `tbl` type to increase the query length for `messages` and `dump1090_terminal`?

(Or do we fork go-sqlite3 and custom compile to remove the 1000 column limit?)

skypuppy: Or, use the much more flexible B+Trees and do away with SQL altogether.

Skypuppy
systems analyst, programmer, database designer and implementer, schema 
designer and implementer, and all-around goofy guy since 1980.

cyoung: Catching up comments to Slack convo - there seems to be quite a bit of diminishing returns in batching a lot of rows, so increasing it to batch insert all rows at once didn't seem to make overall speed faster. At least in my tests, may not be true for all systems. There was *some* benefit to batching. Added code to master now that calculates the max rows based on the `999` column limitation and the number of columns being inserted into the current table.

OK, we might want to roll back some of the logging. Maybe we should implement verbosity levels?

cyoung: @skypuppy - how would this work?
The idea behind this is to correlate all of the data that we have, let it be properly timestamped, and have it work across the multiple hardware profiles we have now.

skypuppy: I just woke up, so if I'm not making sense...

How would what work?  The B-tree method?  The code snippet you post here?

If you mean the B+Tree, you can use the source and compile it into 
stratux and add a record for each timestamp.  If each timestamp data is 
simply output into a text file, one doesn't need any extra code. It can 
all be handled in post-processing which is how I was planning on doing 
it.  One can, of course, use fixed position of each field (lots of empty 
characters sometimes) or one can use field delimiters (commas, quote 
marks, etc) but that incurs even more overhead in the file if most 
fields are populated every time.  To save on SD card space, each 
timestep record (or batch of records) could be compressed in stratux, 
and decompressed during post-processing.  The content of the data 
records will dictate whether that is efficient or not, but for ASCII, it 
usually saves a great deal of disk/SD card space.  Post-processing would 
require a separate executable outside the stratux code itself but that 
is the most efficient method.

I'm a bit confused here, though.  I am at a loss as to the 1,000+ 
columns (fields?) for  each timestamp.  Are there really that many?

Terms: in a 2D database (aka table aka set), columns aka fields are the 
data items that go across, while records aka entries, aka entities go 
down.  As  in one record per timestamp and each record contains the data 
relevant to that, and only that, timestamp. Example: a database (table) 
of stratux users would look something like this --
columns aka fields defined with all ascii text characters as
LastName (10 chars)  FirstName (10 chars) DateStarted  (8 chars) 
TimeStarted (4 chars HHMM) Platform (3 chars code) and so on, yielding 
one row/record, going across the 2D database..
Then, each person is a separate row/record/entity going down the 2D 
database,
SQL has the tables, subtables, and so on, but for  our purposes in this 
example, the set of all stratux users is the total table aka database as 
there is only one "table" in our desired info.  (This example uses easy 
fixed length records just so we can define out terms.)

Let the stratux code just store the timestamp data into a (flat)file, 
and all the heavy lifting is done in post-processing. That way, each 
person can use the database engine they like and they just import that 
data whenever they choose.  Simple.

Timestamps can be unique per record or if there are multiple entries per 
timestamp, non-unique.  If non-unique, we get into 3D, 4D, and so  on 
database, or varying number of entries of each field per timestamp which 
would be an entirely different can of worms.  (I *like* things smple.)  
:)  But then, you'll need a hierarchical database engine instead of a 
relational one.  But that's all the job of the post-processor.

Am I rambling yet?

Did I answer the question?

Skypuppy

skypuppy: Also, are you talking about rolling up ALL records from everyone? Makes 
sense to do that, just requires one extra field.

AvSquirrel:
Closing this, since changes have been manually incorporated as of 8545a9ba0d2d

skypuppy:
AvSquirrel, after line 383, you might insert a failure counter that 
increments each time that happens and save that somewhere where it can 
be reviewed later.  That would help determine if the overhead of sqlite 
is more than the pi is able to keep up with in real-time. Maybe save the 
counter every 15 minutes or so and keep a list  of those values in a 
file for later processing.  Just to be safe.

PR #377, opened 2016-04-08
Better edimax dongle handling in a RPi2 or RPi3

peepsnet:
Since I develop on my Stratux using a second wifi dongle I found that if you plug in an additional wifi dongle into the RPi3 then that dongle will/might be seen as wlan0. If this dongle is an edimax wifi dongle and it is detected as wlan0 then the current stratux-wifi.sh will not be able to start the hostapd service.

This script will now ask the system to give the MAC address of the device at wlan0 and if the system is a RPi2 or RPi3.

If the system is a RPi2/3 and there is an edimax at wlan0 we will now start the hostapd-edimax driver with the hostapd-edimax.conf

This will also solve the issue of a RPi3 user having the on-board wifi stop working and the edimax nano to be used in its place.

Note:

As of right now there are 3 MAC blocks used by Edimax. I have 3 Edimax Nano's and all of them use the "74:da:38" Mac Block so I am not checking for the other 2 MAC blocks at this time. This could be changed in the future. 

http://macaddress.webwat.ch/vendor/Edimax_Technology_Co._Ltd. (add the period back to the end of this link)

I am only guessing that all the Edimax Nano's are in the 74:da:38 range and the 2 others are other devices.

### DIFF
--- a/image/stratux-wifi.sh
+++ b/image/stratux-wifi.sh
@@ -19,7 +19,7 @@ echo 'LSUSB Returned: '$EW7811Un
 RPI_REV=`cat /proc/cpuinfo | grep 'Revision' | awk '{print $3}' | sed 's/^1000//'`
 
 if [ "$RPI_REV" = "a01041" ] || [ "$RPI_REV" = "a21041" ]  || [ "$RPI_REV" = "900092" ]   || [ "$RPI_REV" = "a02082" ]    || [ "$RPI_REV" = "a22082" ]; then
-# If this is a PRPi2/3 [ "$wlan0mac" = '74:da:38' ]
+# If this is a PRPi2/3 
   if [ "$EW7811Un" != '' ] && [[ ${edimaxMac[*]} =~ "$wlan0mac" ]]; then
    # If there is an Edimax Nano at wlan0
    DAEMON_CONF=/etc/hostapd/hostapd-edimax.conf
--- a/image/stratux-wifi.sh
+++ b/image/stratux-wifi.sh
@@ -5,20 +5,26 @@
 /usr/bin/killall -9 hostapd hostapd-edimax
 /usr/sbin/service isc-dhcp-server stop
 
+edimaxMac=(00:1f:1f 80:1f:02 74:da:38)
 
 #  Detect RPi2 or 3 with edimax dongle at wlan0
 #  Per http://elinux.org/RPi_HardwareHistory
 
 DAEMON_CONF=/etc/hostapd/hostapd.conf
 DAEMON_SBIN=/usr/sbin/hostapd
-EW7811Un=$(head -c 8 /sys/class/net/wlan0/address)
-#echo 'WLAN0 MAC '$EW7811Un
+EW7811Un=$(lsusb | grep EW-7811Un)
+wlan0mac=$(head -c 8 /sys/class/net/wlan0/address)
+echo 'WLAN0 MAC: '$wlan0mac
+echo 'LSUSB Returned: '$EW7811Un
 RPI_REV=`cat /proc/cpuinfo | grep 'Revision' | awk '{print $3}' | sed 's/^1000//'`
 
-if [ "$RPI_REV" = "a01041" ] || [ "$RPI_REV" = "a21041" ]  || [ "$RPI_REV" = "900092" ]   || [ "$RPI_REV" = "a02082" ]    || [ "$RPI_REV" = "a22082" ]  &&   [ "$EW7811Un" = '74:da:38' ]; then
- # This is a RPi2B or RPi3B with Edimax Nano at wlan0
- DAEMON_CONF=/etc/hostapd/hostapd-edimax.conf
- DAEMON_SBIN=/usr/sbin/hostapd-edimax
+if [ "$RPI_REV" = "a01041" ] || [ "$RPI_REV" = "a21041" ]  || [ "$RPI_REV" = "900092" ]   || [ "$RPI_REV" = "a02082" ]    || [ "$RPI_REV" = "a22082" ]; then
+# If this is a PRPi2/3 [ "$wlan0mac" = '74:da:38' ]
+  if [ "$EW7811Un" != '' ] && [[ ${edimaxMac[*]} =~ "$wlan0mac" ]]; then
+   # If there is an Edimax Nano at wlan0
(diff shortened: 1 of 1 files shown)


jpoirier:
LGTM

JohnOCFII:
What is the desired WiFi adapter selection logic?

If there is a built-in WiFi adapter, use that for Stratux?
If there is an external WiFi adapter (any brand?) use that for Stratux?

What you have written seems to make sense, but not sure if it catches all the likely cases.  Also - at some point, would you want the choice of WiFi adapter to be a user choice in the WebUI?

peepsnet:
Well its not that simple. 
There are a few files that are coded to use wlan0 as the hotspot. 
These files would need to be modified

Example:
https://github.com/cyoung/stratux/blob/master/image/isc-dhcp-server
'/etc/default/isc-dhcp-server' in your stratux
uses the line:
INTERFACES="wlan0"

and

https://github.com/cyoung/stratux/blob/master/image/interfaces
'/etc/network'intervaces' in your stratux
uses the lines
iface wlan0 inet static
  address 192.168.10.1
  netmask 255.255.255.0
  post-up /usr/sbin/stratux-wifi.sh

So both of these files are set to use wlan0 as the hostapd. 

I contemplated 2 approaches to this. 
First: Change/rewrite the 2 files above to use wlan1 if you are using a RPi3 w/edimax detected at wlan0

Second: keep the files and just change the driver that wlan0 is using if an edimax nano is found at wlan0 and any RPi2/3 is found(this consolidates the RPi2 and 3 logic also). So this would make the onboard wifi not the hotspot but who cares. You plugged in a second dongle and you just want it to work.)

Also with this option. I person having issue with the onboard wifi adaptor(it quit working/need more power/testing reception) can simply add another dongle and improve the wifi. 

Option 2 seemed to be the easier and most universal solution. 

With option one(in my setup, so I would develop for it) I would have to not only change wlan0 to wlan1 but modify other things associated with that second connection to link to my home WIFI. 

So in summery:
Stratux is, right now, built to always use wlan0(what ever the system assigns to wlano) as the hotspot. If an edimax nano is plugged in to a RPi2/3 and it is detected as wlan0 then we need to use the proprietary hostapd driver so why not catch the error and prevent it.

jpoirier:
On Fri, Apr 8, 2016 at 8:38 AM, JohnOCFII <notifications@github.com> wrote:

> What is the desired WiFi adapter selection logic?
>
> If there is a built-in WiFi adapter, use that for Stratux?
> If there is an external WiFi adapter (any brand?) use that for Stratux?
>
> What you have written seems to make sense, but not sure if it catches all
> the likely cases. Also - at some point, would you want the choice of WiFi
> adapter to be a user choice in the WebUI?
>
The main use case for stratux is as a dedicated unit and, afaik, the only
supported config is a single wifi device. This modification is meant more
for convenience, with zero side effects to the current functionality, than
it is to add additional user config options.

peepsnet:
Better said @jamez70

skypuppy:
Would it not be easier to change the onboard wifi to always be configured as wlan0?  I vaguely remember we can do that with ifconfig (or ipconfig?) at the startup of stratux or upon boot?

jamez70:
There are a few wifi adapters I was able to get to work with the Stratux. THe TP Link TL-WN725N is one I was able to get at a local store (Micro Center) and it works very well. When I built my first one I did not have a wifi adapter and bought a belkin N300 and by pure luck it worked too.

I read somewhere the RPI 3's Wifi adapter was not capable of being an access point. Is this not true?

JohnOCFII:
> I read somewhere the RPI 3's Wifi adapter was not capable of being an access point. Is this not true?

The RPi 3 WiFi works fine as an Access Point in Stratux.

peepsnet:
@skypuppy 

Configuring the onboard as the hotspot is involved. if you look at my 2 options you can see I choose the easier

skypuppy:
Might be easier for the moment but could get wicked with all the if-thens of configuring the "other" wifi reliably.

peepsnet:
its just a one off issue... i use 2 wifi's  but if for some reason the pi3 onboard wifi broke this wold allow for a user to use an edimax dongle aslo...  and if you read the latest on reddit... We are trying to help a guy who said his edimax in the pi2 didnt work but he got a pi3 and plugged in the edimax and it worked...  which this fix proves is not possible... 

So right after i look at this error someone says they plugged in their edimax into their pi3 because he is misunderstanding something... 

lol

cyoung:
`/etc/udev/rules.d/70-persistent-net.rules` ?

peepsnet:
That file does not exist on my pi3

peepsnet:
Are you suggesting create the file to fix this issue??

peepsnet:
https://rundata.wordpress.com/2013/11/20/renaming-network-interfaces/ ??

cyoung:
```
SUBSYSTEM=="net", ACTION=="add", DRIVERS=="rtl8192cu", ATTR{address}=="?*", ATTR{dev_id}=="0x0", ATTR{type}=="1", KERNEL=="wlan*", NAME="wlan10"
```

Works alright... but you're using two EDIMAX at once?

cyoung:
Try this, it will put the on board Pi3 Wi-Fi on `wlan0` always if available. Could you check on a Pi2 if it interferes with anything?
```
SUBSYSTEM=="net", ACTION=="add", DRIVERS=="brcmfmac_sdio", ATTR{address}=="?*", ATTR{dev_id}=="0x0", ATTR{type}=="1", KERNEL=="wlan*", NAME="wlan0"
```

I think that's all we need across all installations.

peepsnet:
Sorry for the delay. This just changed the name of the device to `rename4`

I was not able to get this to work.

tonymorris:
In an attempt to have a stratux (rpi3) with a second network interface (acting as client), I have successfully applied a custom variation of this script.

https://github.com/peepsnet/stratux/blob/master/image/stratuxPi3EdimaxFix.sh

----

Original thread.
https://www.reddit.com/r/stratux/comments/4m6iwt/second_wifi_adapter/

peepsnet:
I just hooked my edimax to my RPI3 with Version: v1.0r1 (5a13cc01ee) and the onboard WIFI is wlan is detected as WLAN0 so this code may be obsolete at this time!

PR #372, opened 2016-04-07
Add mutex to dataLogTimestamps[]

AvSquirrel:
**Problem:**
Intermittent out-of-bounds references to `dataLogTimestamps[dataLogCurTimestamp] = ts` cause crash after unpredictable amount of time. Cause seems to be simultaneous additions of new values to `dataLogTimestamps` in  `checkTimestamp()` and `setDataLogTimeWithGPS()`. 
```
unexpected fault address 0x30303030
fatal error: fault
[signal 0xb code=0x2 addr=0x30303030 pc=0xd7da8]

goroutine 45 [running]:
runtime.throw(0x4a0238, 0x5)
        /root/go/src/runtime/panic.go:527 +0x78 fp=0x10b31430 sp=0x10b31424
runtime.sigpanic()
        /root/go/src/runtime/sigpanic_unix.go:27 +0x280 fp=0x10b3145c sp=0x10b31430
runtime.evacuate(0x3f7550, 0x108c8260, 0x274)
        /root/go/src/runtime/hashmap.go:819 +0x1b8 fp=0x10b314c0 sp=0x10b31460
runtime.growWork(0x3f7550, 0x108c8260, 0x674)
        /root/go/src/runtime/hashmap.go:791 +0x6c fp=0x10b314d0 sp=0x10b314c0
runtime.mapassign1(0x3f7550, 0x108c8260, 0x67db90, 0x10b31574)
        /root/go/src/runtime/hashmap.go:433 +0x14c fp=0x10b31524 sp=0x10b314d0
main.setDataLogTimeWithGPS(0x1080f250, 0x4787d680, 0x42338aef, 0xc2bad45a, 0x1, 0x445228bd, 0xc2c9718f, 0xf0008, 0xa, 0x40f33333, ...)
        /root/stratux/main/datalog.go:441 +0x110 fp=0x10b315b0 sp=0x10b31524
main.processNMEALine(0x11c67d60, 0x42, 0x0)
        /root/stratux/main/ry835ai.go:678 +0x1c08 fp=0x10b31ebc sp=0x10b315b0
```

**Solution:**
Protect `dataLogCurTimestamp++` and `dataLogTimestamps[dataLogCurTimestamp]` by wrapping portions of `checkTimestamp()` and `setDataLogTimeWithGPS()` with mutex locks.

PR #386, opened 2016-04-10
SQLite logging - Change dataLogTimestamps[] to slice

AvSquirrel:
Referenced in #372.

`dataLogTimestamps` is a golang map. The map type is not threadsafe, and requires synchronization of all writers and readers. In this application, not providing mutex sync around all readers still allowed gen_gdl90 to randomly crash, and adding full protection caused too many deadlocks.

This PR changes `dataLogTimestamps` to a more conventional array / slice. This makes more sense for how it's being used (storing a series of sequential timestamp structures). It doesn't need synchronization if we're not removing values, and appends seem to be faster than map inserts -- this speeds up the SQLite logging.

### DIFF
--- a/main/datalog.go
+++ b/main/datalog.go
@@ -20,7 +20,7 @@ import (
 	"reflect"
 	"strconv"
 	"strings"
-	"sync"
+	//	"sync"
 	"time"
 )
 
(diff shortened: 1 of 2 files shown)


skypuppy:
AvSquirrel, have you considered having only a single function (with a sizable queue that can grow and shrink by, say 50kb [array of pointers],  on demand) whose purpose is writing to the file?  All other functions/processes feed that single writer?  No mutex overhead but on the Pi I don't know whether inter-process messages (D-bus?) is cheaper or more expensive than mutexes.

Then again, it might be better, overall, to incorporate dump1090 into the main code, as is dump978, giving much less OS overhead.

cyoung:
Good call.

PR #387, opened 2016-04-12
typo

peepsnet:


cyoung:
Ok, but the install puts it at `/root/.stxAliases`. I'll just fix it the other way around.

PR #340, opened 2016-03-17
GPS, AHRS, and traffic enhancements. ForeFlight compatibility improvements.

AvSquirrel:
- Add equation library with functions for local and great circle distance / bearing, linear regression, weighted linear regression, and other statistics functions
- Improved GPS detection / validation #253
- Add support for MTK3339 GPS autoconfiguration on GPIO #315 (Testing; needs validation)
- GPS pseudo-attitude calculations for setups without hardware AHRS #136 (Testing; needs validation on non-ublox hardware and <5 Hz sample rates)
- Allow disabling AHRS GDL90 output without disabling AHRS hardware #332 (ForeFlight 7.6 workaround)
- Add support for FLARM-formatted traffic output over ttyUSB0 #218 (Testing; needs validation)
- Add support for ownship / traffic / AHRS output to ForeFlight using "flight simulator" interface (Testing; intended for developer use; may be limited to ground use only)
- Improved traffic UI page, showing distance / bearing to target, and non-position Mode S traffic #294
- CSV logging of traffic data #294
- Add parsing of ownship ADS-B / Mode S pressure altitude and callsign.
- Separated verbose debug logging from prepended traffic source option
- Bug fixes
- WebUI documentation updates

PR #390, opened 2016-04-13
Traffic backend updates

AvSquirrel:
This PR covers a few improvements to the traffic backend, and will be the first of a few smaller PRs referenced in yesterday's comments on #340:

- Updated dump1090 to include latest commits to the main dump1090-mutability fork
- Add equations.go library for great circle distance / bearing (and eventual use for GPS pseudoattitude / AHRS improvements)
- Calculate great circle distance and bearing to traffic targets
- Add squawk codes for Mode S targets
- Move SQLite logging of the raw port 30006 dump1090 JSON messages to a new table
- SQLite logging of dump1090 console only when globalSettings.DEBUG = true and recordLogs = true to save disk space
- SQLite logging of decoded traffic information reduced to once per second, when each `ti` is sent to the GDL90 output. This will cut down on table size.
- Improved method for prepending traffic source to callsigns. (TODO: create a new globalSettings variable to separate the callsign tagging from debug logging)
- Improvements for `demoTraffic` debugging routine

This PR also *excludes* a couple of new features that were included in #340:
- CSV output of traffic information. This is redundant now that we have SQLite output
- FLARM formatted traffic output (may be included in a separate branch at some point)
- XTRAFFIC formatted traffic output for the "ForeFlight AHRS" mode (per @cyoung, mainly a developer tool; may be included in a separate branch at some point)

@cyoung , my rationale for using `logESMsg(eslog) ` rather than `logMsg(thisMsg)` is:

1. Saving a little space in the database (and in the INSERTs) by eliminating overhead involved in writing unused fields (signal strength, tower ID, ADS-B products, etc.)  and 
2. More critically, the `sqlTypeMap` in datalog.go considers `[]byte` an unsupported type, and therefore `msg.Data` is not written to the DB. 

The latter is more of a general bug in the SQLite logging (i.e. I expect UAT data is affected as well, but I haven't run this code in earshot of a tower to confirm) but I wanted to get your thoughts on whether it's a reasonable approach to separate the raw UAT and ES logs, and whether this approach makes sense.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ all:
 
 xgen_gdl90:
 	go get -t -d -v ./main ./test ./linux-mpu9150/mpu ./godump978 ./mpu6050 ./uatparse
-	go build $(BUILDINFO) -p 4 main/gen_gdl90.go main/traffic.go main/ry835ai.go main/network.go main/managementinterface.go main/sdr.go main/uibroadcast.go main/monotonic.go main/datalog.go
+	go build $(BUILDINFO) -p 4 main/gen_gdl90.go main/traffic.go main/ry835ai.go main/network.go main/managementinterface.go main/sdr.go main/uibroadcast.go main/monotonic.go main/datalog.go main/equations.go
 
 xdump1090:
 	git submodule update --init
--- a/main/datalog.go
+++ b/main/datalog.go
@@ -447,6 +447,7 @@ func dataLog() {
 		makeTable(globalSettings, "settings", db)
 		makeTable(TrafficInfo{}, "traffic", db)
 		makeTable(msg{}, "messages", db)
+		makeTable(esmsg{}, "es_messages", db)
 		makeTable(Dump1090TermMessage{}, "dump1090_terminal", db)
 	}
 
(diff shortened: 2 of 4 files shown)


cyoung:
What are the changes on dump1090?

AvSquirrel:
Added documentation for the Stratux port 30006 output, and brought it up to date with other changes / improvements on the upstream fork. 

[Here's the diff.](https://github.com/AvSquirrel/dump1090/compare/9a4fb85...AvSquirrel:master)

The only change that will affect Stratux is that messages with 2-bit error correction will no longer be sent on the port 30006 output. This will reduce the possibility of bringing in junk data, at the cost of losing a few messages from very weak signals.

cyoung:
Okay, well the version update only shows changes on the mutability branch which I don't care to review now. I'd rather not constantly update to the very latest dump1090 unless there is a real reason to do so. So this will be rolled back after the merge.

Are `Age` and `AgeLastAlt` used anywhere? I see they are updated in `sendTrafficUpdates()` but the values not actually used.

AvSquirrel:
> Are `Age` and `AgeLastAlt` used anywhere? I see they are updated in `sendTrafficUpdates()` but the values not actually used.

In the traffic UI, both as an indicator of traffic age for the user, and to time-out and remove stale data from the page: https://github.com/AvSquirrel/stratux/commit/d00ab538cd20c81aed884a9c7bf280277bb6c4c7

(That branch, with a couple of formatting changes, will be the basis of my traffic UI improvements... I'll submit a PR after you close #390)

PR #408, opened 2016-04-30
GPS backend improvements

AvSquirrel:
A few incremental improvements to GPS subsystem:
- Prioritize `/dev/ttyACM0` over `/dev/ttyUSB0` at startup. This will allow the simultaneous use of u-blox receivers over their native USB connection, while having a USB-to-serial bridges installed.
- 'isGPSValid()' now checks for a connected GPS device, and that the "quality" byte is greater than zero. This should prevent invalid positions (e.g. off the coast of Equatorial Guinea, 0°N / 0°W) from being used for traffic distance calculations at startup.
- Changed UI description for `quality==2` solution from "DGPS (SBAS / WAAS)" to "GPS + SBAS (WAAS / EGNOS)". While satellite-based augmentation is a differential method, "DGPS" more commonly refers to terrestrial differential corrections. The new description is more accurate for 99.9% of Stratux use cases. (The [EGNOS](https://en.wikipedia.org/wiki/European_Geostationary_Navigation_Overlay_Service) label is being added as a tip-of-the-hat to European Stratux users. It is Europe's SBAS, and serves the same role as WAAS in North America.)

### DIFF
--- a/main/gen_gdl90.go
+++ b/main/gen_gdl90.go
@@ -721,7 +721,7 @@ func cpuTempMonitor() {
 
 func updateStatus() {
 	if mySituation.Quality == 2 {
-		globalStatus.GPS_solution = "DGPS (SBAS / WAAS)"
+		globalStatus.GPS_solution = "GPS + SBAS (WAAS / EGNOS)"
 	} else if mySituation.Quality == 1 {
 		globalStatus.GPS_solution = "3D GPS"
 	} else if mySituation.Quality == 6 {
--- a/main/ry835ai.go
+++ b/main/ry835ai.go
@@ -130,15 +130,13 @@ func initGPSSerial() bool {
 	baudrate := int(9600)
 	isSirfIV := bool(false)
 
-	if _, err := os.Stat("/dev/ttyUSB0"); err == nil {
+	if _, err := os.Stat("/dev/ttyACM0"); err == nil { // u-blox receivers on native USB connection
+		device = "/dev/ttyACM0"
+	} else if _, err := os.Stat("/dev/ttyUSB0"); err == nil { // USB-to-serial bridge. Typical use is BU-353-S4 SIRF IV receivers, but could also be for other devices or serial-out (better detection is TODO)
 		isSirfIV = true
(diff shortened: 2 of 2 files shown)


skypuppy:
A small nit:  doesn't quality have to be '3' to indicate SBAS fix?  I am *really* tired so I may be incorrect.

AvSquirrel:
> A small nit: doesn't quality have to be '3' to indicate SBAS fix? I am really tired so I may be incorrect.

This is the quality byte (field no. 6) in the GGA messages:
- 0 == no fix.
- 1 == GNSS (GPS, GLONASS, etc.).
- 2 == GNSS + differential corrections (DGPS, SBAS, etc.).
- 6 == dead reckoning (for receivers that incorporate gyros, speedometers, or other sensors to predict position after a loss of satellite lock... it's not used by any of the receivers anyone's using on this project).

Maybe you're thinking of the GSA messages? Those have a "navMode" field where 1 == no fix, 2 == 2D fix, and 3 == 3D fix.

cyoung:
We really need to separate out the different elements of `mySituation` and find when certain parts are changing and when others aren't. It's a bit uncomfortable to see select variables being updated in the fix data but not others. For example, why `mySituation.Satellites = 0` and not also `mySituation.SatellitesTracked = 0` and `mySituation.SatellitesSeen = 0`?

AvSquirrel:
> For example, why mySituation.Satellites = 0 and not also mySituation.SatellitesTracked = 0 and mySituation.SatellitesSeen = 0?

`SatellitesTracked` and `SateliitesSeen` are different metrics, provided by different NMEA messages (PUBX,03 or GxGSV). Getting a better handle on those from non-Ublox receivers is next on my to-do list.

Imagine an airplane that's just been pulled into a hangar. The GPS could have 15 tracking channels open looking for satellites it "knows" are above the horizon based on its last known position and almanac data (`SatellitesTracked`), only be receiving signals from three due to the shielding effects of the hangar (`SateliitesSeen`), and therefore not be able to calculate a position fix ("Satellites" = 0).

PR #410, opened 2016-05-03
Graceful shutdown of SQLite logs

AvSquirrel:
- Add watchdog function and quitchannel controls for graceful shutdown and restart of sqlite logs. Log may now be started and stopped at any time.
- SQLite system now only runs when `globalSettings.ReplayLog == true`.
- SQLite database is safely and properly closed when log system is closed.
- globalStatus and globalSettings are only sent to log when `globalSettings.ReplayLog == true`
- Display of traffic source in callsign now split out from verbose 'DEBUG' setting

### DIFF
--- a/main/datalog.go
+++ b/main/datalog.go
@@ -377,9 +377,9 @@ func dataLogWriter(db *sql.DB) {
 			//			}
 			timeStart := stratuxClock.Time
 			nRows := len(rowsQueuedForWrite)
-			//if globalSettings.DEBUG {
-			log.Printf("Writing %d rows\n", nRows)
-			//}
+			if globalSettings.DEBUG {
+				log.Printf("Writing %d rows\n", nRows)
+			}
 			// Write the buffered rows. This will block while it is writing.
 			// Save the names of the tables affected so that we can run bulkInsert() on after the insertData() calls.
 			tblsAffected := make(map[string]bool)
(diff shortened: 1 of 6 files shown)


cyoung:
Thanks, we've needed `DisplayTrafficSource` for quite some time.
When `ReplayLog` is off, the sqlite log is just status/timestamp updates. Does it get really that large that it's an issue for normal users?

cyoung:
Alright, we'll let sqlite stay off completely for now by default until we've established that it's safe at 100% logging in different use cases.

AvSquirrel:
I made a build that only enabled sqlite logging of status and timestamp updates, and let it run overnight after deleting stratux.log and stratux.sqlite. With 978/1090 and GPS enabled and  "Verbose Message Log" disabled, stratux.sqlite was 758 KB, and stratux.log was 1641 KB after 19.6 hours of runtime.

That's roughly 3 MB per day, or just over 1 GB per year running continuously. So...

> Does it get really that large that it's an issue for normal users?

... probably not, at least at this stage in development. A novice user who doesn't expand the file system will have an unusable Stratux after about 10,000 (give or take) continuous power-on hours.

My concerns were more about compatibility with RO filesystems, and a better ability to manage log files (including closing the sqlite file on exit / on demand).

PR #412, opened 2016-05-09
GPS improvements: Satellite info; multi-GNSS support; UI updates

AvSquirrel:
This update relates to backend and UI improvements of the GPS system:

- Statistics for individual satellites (identification, elevation, azimuth, signal level, and a 'used in solution' flag) are now parsed from PUBX03, GPGSV, and GLGSV messages. GPS, SBAS, and GLONASS are all supported.
- Data for individual satellites are stored in a `map` structure, similar to the traffic interface. This allows for calculation of statistics for the entire GNSS constellation.
- Multi-GNSS (GLONASS + GPS) now supported by default for u-blox 8 receivers (e.g. RY835/RY836). This will be ignored by u-blox 7 receivers (VK-172). Default sampling rate is now 5 Hz for all supported (VK-172, RY83x, BU-353-S4) and semi-supported (Adafruit Ultimate) GPS receivers.
- SIRF IV configuration for BU-353-S4 has been updated to output GSV messages
- Creation of new `http://192.168.10.1/getSatellites` JSON interface to make data available to external applications and web UI
- Updated 'GPS/AHRS' UI page to show the satellite table
- Updated help file for GPS UI page to describe new features

Notes: My RY835AI and RY836AI both shipped from Reyax with firmware version 2.01. This limits the PUBX,03 satellite message to no more than 20 entries. This causes higher-numbered GLONASS satellites to be excluded from the data table. **I recommend updating to version 3.01. The [firmware update](https://www.u-blox.com/sites/default/files/UBX_M8_301_SPG.911f2b77b649eb90f4be14ce56717b49.bin) can be installed with the u-blox "u-center" software to remove this limitation.**

**Screenshots**
WAAS solution on an RY836AI over serial. GLONASS satellites are being tracked and received, but not used in the solution. (GLONASS is excluded from the position solution once WAAS corrections are applied.)
image

GPS-only solution on a BU-353-S4. WAAS satellite 133 is in view, but differential corrections haven't been applied yet.
image


main/managementinterface.go
```diff
@@ -160,6 +160,17 @@ func handleTowersRequest(w http.ResponseWriter, r *http.Request) {
 	fmt.Fprintf(w, "%s\n", towersJSON)
 }
 
+// AJAX call - /getSatellites. Responds with all GNSS satellites that are being tracked, along with status information.
+func handleSatellitesRequest(w http.ResponseWriter, r *http.Request) {
+	setNoCache(w)
+	setJSONHeaders(w)
+	satellitesJSON, err := json.Marshal(&Satellites)
```
cyoung: Needs mutex.

main/ry835ai.go
```diff
@@ -283,9 +316,8 @@ func initGPSSerial() bool {
 		p.Write(makeUBXCFG(0x06, 0x01, 8, []byte{0xF0, 0x0D, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00})) // GNS
 		p.Write(makeUBXCFG(0x06, 0x01, 8, []byte{0xF0, 0x0E, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00})) // ???
 		p.Write(makeUBXCFG(0x06, 0x01, 8, []byte{0xF0, 0x0F, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00})) // VLW
-
 		p.Write(makeUBXCFG(0x06, 0x01, 8, []byte{0xF1, 0x00, 0x01, 0x01, 0x01, 0x01, 0x01, 0x00})) // Ublox,0
-		p.Write(makeUBXCFG(0x06, 0x01, 8, []byte{0xF1, 0x03, 0x0A, 0x0A, 0x0A, 0x0A, 0x0A, 0x00})) // Ublox,3
+		p.Write(makeUBXCFG(0x06, 0x01, 8, []byte{0xF1, 0x03, 0x05, 0x05, 0x05, 0x05, 0x05, 0x00})) // Ublox,3
```
cyoung: What's the change here?

main/ry835ai.go
```diff
@@ -586,44 +618,131 @@ func processNMEALine(l string) (sentenceUsed bool) {
...
 			// We've made it this far, so that means we've processed "everything" and can now make the change to mySituation.
 			mySituation = tmpSituation
 			return true
-		} else if x[1] == "03" { // satellite status message
+		} else if x[1] == "03" { // satellite status message. Only the first 20 satellites will be reported in this message for UBX firmware older than v3.0. Order seems to be GPS, then SBAS, then GLONASS.
 
 			// field 2 = number of satellites tracked
-			satSeen := 0 // satellites seen (signal present)
+			//satSeen := 0 // satellites seen (signal present)
 			satTracked, err := strconv.Atoi(x[2])
```
cyoung: Shouldn't rely on this, better to do as below based on the length of the actual message - in the case of a truncated message the below `for {}` block will cause an index overflow crash.

main/ry835ai.go
```diff
@@ -1160,6 +1436,39 @@ func attitudeReaderSender() {
 	globalStatus.RY835AI_connected = false
 }
 
+/*
+	updateContellation(): Periodic cleanup and statistics calculation for 'Satellites'
+		data structure. Calling functions must protect this in a satelliteMutex.
```
cyoung: 👍

main/managementinterface.go
AvSquirrel: Good catch. Will fix and push a commit.

main/ry835ai.go
AvSquirrel: I'm reducing the sample frequency from 10 Hz to 5 Hz (see line 262), but I still want satellite status to update once a second. So, I'm setting the PUBX,03 message to send every 5th position (0x05) instead of every 10th position (0x0A).

main/ry835ai.go
AvSquirrel: Ah, sure. We really shouldn't see that situation since a truncated message should fail NMEA checksum, but it's easy to test for.

Unlike the GSA message you're referring to, PUBX,03 has a variable number of fields. I'll do one check to see if the length is at least three so "number of satellites" can be parsed, and then calculate the expected total length based on that.

cyoung: If there's a one byte corruption in the NMEA sentence on the ",", then there is a 1/255 chance that it both passes the checksum and the message is parsed as too short - thus causing a fatal crash.
I don't know what the probability of getting corrupted messages is, but it seems high enough because of how often we see it in the logs.

AvSquirrel: fc1870a

main/managementinterface.go
AvSquirrel: This was addressed with	f02fdf1

AvSquirrel:
@cyoung -- commit a39a870 is a minor bug fix to allow SBAS satellites to be shown as "in solution" for receivers that use GSA / GSV messages (BU-353-S4, MTK3339) even if the GSA message does not list SBAS satellites.

The stability checks you suggested (mutex aroud the AJAX handler and index checking for the PUBX,03 messages) were previously committed to this branch as f02fdf1 and fc1870a.

Can you review again and merge if these look OK?

cyoung:
Looking good, thanks!

Issue #420, opened 2016-05-20
Loss of Foreflight and UI connection in flight


cbjohnsonjr:
1. Stratux version: 0.8r2
2. Stratux config: Raspberry Pi 3, 1 SDR Nano 2( 978 only), no GPS, no AHRS, Battery - Anker® 2nd Gen Astro E5 16000mAh, short micro usb cable
3. EFB app and version : Foreflight v7.6.1 and Flightplan same issue
4. Description of your issue : Loss of Forelight connection and UI connection  occurs every flight with duration approximately 1.0 - 1.5 hours. Jet aircraft operating FL 350-450. I unplug power then repower  and device operates normally for a few minutes then problem reoccurs. If I leave power off a for longer period then last operates for 1.0+ hour then problem starts again. I've tried different batteries,  full sized receiver, placing receiver in cooler spot (typical reporting temp 50 degrees C), all no change. Wifi never goes away only EFB and UI connection. On the ground with no ADSB tower connection continues to operate without problems. Everything else works

peepsnet:
Did you turn AHRS off in the webUI???

https://github.com/cyoung/stratux/issues/332

Ergonomicmike:
Sounds like a classic overheating problem. But since you put it in a cooler spot ... Maybe try a different SD card in case it's failing with a little heat? (No ADS-B on ground, things don't get as hot.)

idriss-ophawk:
I have the same issue with:
- Rasberry Pi 2
- Stratux version: 0.8r2 and 0.8r1
The stratux is in a ventilated area: CPUTemp around 31.48 deg C. 

It started to appear after I bought a more powerful antenna. 
I now get 6-10 UAT towers but my stratux freezes after 7-10 minutes. 
By freeze I mean that my ipad is still connected to startux in wifi but the control center and all other operations are not available (GPS, UAT, etc.).
I have to unplug/plug it to restart it and it works again. I can't get it work longer than 10min in the air. 
On the ground, it lasts longer (30min+).

I enabled the logs in flight, the file is attached without the "Invalid NMEA string: Invalid NMEA message" entries which are cluttering the file but don't seem to be the issue. 

[stratux-clean.txt](https://github.com/cyoung/stratux/files/276595/stratux-clean.txt)

I also have the UAT and GPS log if needed.

cbjohnsonjr:
[...]

PR #428, opened 2016-05-27
Additional UAT stability improvements (tower mutex, array bounds checks)

AvSquirrel:
This is a rollup of stability improvements tested in conjunction with #420. Includes:

- Mutex protection for an unprotected `map[string]ADSBTower `
- Removes a currently unused `products_last_minute` map type
- Array bounds checking on uatdecode.go `DecodeUplink()` function
- Array bounds checking on the (now disabled) `decodeAirmet()` function

### DIFF
--- a/main/gen_gdl90.go
+++ b/main/gen_gdl90.go
@@ -26,6 +26,7 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
+	"sync"
 	"syscall"
 	"time"
 
@@ -125,6 +126,7 @@ type ADSBTower struct {
 }
 
 var ADSBTowers map[string]ADSBTower // Running list of all towers seen. (lat,lng) -> ADSBTower
+var ADSBTowerMutex *sync.Mutex
 
 func constructFilenames() {
 	var fileIndexNumber uint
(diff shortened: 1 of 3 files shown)


cyoung:
1. `ADSBTowerMutex` missed in `parseInput()`.
2. Will just remove the `products_last_minute` code if we don't need it, otherwise we'll have the code cluttered with a bunch of commented code.

Thanks!

cyoung:
What do you think about showing products breakdown in recent UAT messages?

https://github.com/cyoung/stratux/blob/master/web/plates/status.html#L51-L61

`MsgLog.Products` isn't used, either. It was displaying at one point. Not all EFBs show this breakdown and there have been a few questions "am I receiving NEXRAD frames?" etc.

ssokol:
I'm all in favor of adding the product count. I regularly get reports from
users claiming that they're not getting any NEXRAD data. It would be great
if there was a status item that showed what had come in.

On Tue, May 31, 2016 at 7:32 AM, cyoung <notifications@github.com> wrote:

> What do you think about showing products breakdown in recent UAT messages?
>
>
> https://github.com/cyoung/stratux/blob/master/web/plates/status.html#L51-L61
>
> MsgLog.Products isn't used, either. It was displaying at one point. Not
> all EFBs show this breakdown and there have been a few questions "am I
> receiving NEXRAD frames?" etc.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/cyoung/stratux/pull/428#issuecomment-222674512>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AAVYPHob24tlcvi5YRA5WyCi7v5VDxPbks5qHCpagaJpZM4Io6yG>
> .
>



-- 
Steven Sokol
408 Camelot Drive
Liberty, MO 64068

mobile: +1 816-806-8844
fax: +1 816-817-0441

PR #430, opened 2016-05-29
Add support for RPi0 board rev 1.3

jpoirier:
Adds a serial check for the board released second quarter of 2016


cyoung:
Thank you!

PR #438, opened 2016-06-08
Bug fix for logging

AvSquirrel:
Commit 404a4c9c0 introduced a bug that broke logging for non-FlightBox hardware.

If FlightBox wasn't detected, `debugLogf` `dataLogFilef` were never assigned and remained as nulls. Consequently, the SQLite log and message log were never written to disk, and an error was thrown on the UI status page.


cyoung:
Good catch, thank you.

Issue #426, opened 2016-05-27
Remove bloat from interface pages


cyoung:
1. Stratux version: 47271df7647b4efa49d86409d98897400559b018
2. Stratux config:
   
   SDR
     [x] single
     [ ] dual
   
   GPS
     [ ] yes
     [x] no
     type:
   
   AHRS
     [ ] yes
     [x] no
   
   power source: wall
   
   usb cable: a good one
3. EFB app and version: N/A
   
   EFB platform: N/A
   
   EFB hardware: N/A
4. Description of your issue:

a. I have GPS disabled but it's showing GPS-enabled options on the traffic page.

![a](https://cloud.githubusercontent.com/assets/57193/15596600/de64d0c2-2396-11e6-9402-701860a2b375.png)

b. What is the clock difference used for now? We've abstracted away from RTC dependence.

![b](https://cloud.githubusercontent.com/assets/57193/15596620/fe5eaa74-2396-11e6-93be-d9252c148352.png)

Ergonomicmike:
On the clock difference: I was kind of shocked to see my clock dif be some huge negative number the other day when I started the Stratux. Turns out that it had been more than a few days since I had powered up the BU-353 and it took a minute to get GPS time. So the clock dif could be considered a Diagnostic value if something goes wrong with the GPS.

As for abstracting away from Pi time, if one is not using GPS on the Stratux, then one is stuck on Pi time. Does that matter for anything else? It makes the time stamp in the logs goofy. Any implications for traffic or FIS-B? Should there be a way for those users to capture their device time and paste it to the Pi?

AMNegron:
Do UAT messages send a time stamp? If not, should there be a way to manually update the clock? And as Ergonomicmike asked, is it required?

AvSquirrel:
@cyoung - 

For a while, some power users ran into problems with certificate issues if they tried to pull and build on a Pi with incorrect clock settings. The move to .sh updates mitigates that. Developers will just have to know to use a GPS, set the clock manually, or enable ntp.

With all of the user-facing functions using the monotonic ticker, about the only reason I can think to keep user-facing time around is log correlation, as @Ergonomicmike notes. I wouldn't call it a strong reason to keep it.


@AMNegron -- traffic messages are flagged with the timeslice within the second they are sent, but are not timestamped. Some of the FIS-B (weather) messages have timestamps / datestamps for message applicability (e.g. NOTAMs), but not the actual message send time.

Issue #136, opened 2015-12-08
Wanted: GPS pseudo-attitude calculations


bradanlane:
Anyone happen to have stumbled across code that does GPS pseudo-attitude calculations? Thus far, all I can find are academic papers on the subject and they can be tough to decipher.

GPS pseudo-attitude has been around for more than 15 years and is how Garmin and other vendors depict a "six pack" on a portable GPS.

I'd like to tinker with implementing a GPS based version of @bagarre Pi2D2.

Nokomis449:
Not sure if this will help as it's way over my head, but check out the source code they mention here:
http://diydrones.com/profiles/blogs/705844:BlogPost:16708
It uses AHRS as well as GPS and I think you're looking for an all-GPS solution, but this code is supposedly well documented so maybe it can give some ideas.

AvSquirrel:
Thinking about the Garmin six-pack, we're halfway there.

Airspeed: Use GPS ground speed.
Altimeter: Use GPS altitude.
Heading indicator: Use GPS track.

For the rest of the instruments, turn coordinator and VSI are straightforward. Pseudo-attitude would be a function of those two. The challenge, if there is one, is ensuring that the GPS data is fast enough and noise-free enough to have useable inputs. I'll look at writing and testing an implementation this weekend.

**Turn coordinator:** Turn rate is time derivative of heading. GPS heading is stable and we want to react quickly to changes, so I'd calculate the slope of heading vs time over no more than 3-5 points... so two seconds (assuming 1 Hz GPS sampling), or the last second (if sample rate is 5 Hz or higher). 

**VSI:** Vertical speed is time derivative of altitude. GPS altitude (especially non-WAAS) can drift a few meters around the actual altitude, so filtering would be needed, especially since this will be fed into the pseudo pitch calculation. I'd try to use at least 5, preferably 10 points depending on GPS refresh rate. Report only if GS is greater than 30 knots (?) 

**Attitude indicator - pitch:** Approximate as the flight path, so this simplifies to atan(VS/GS) as long as vertical speed and ground speed are converted to the same units. Report only if GS is greater than 30 (?) knots.

**Attitude indicator - roll:** Assuming coordinated flight, turn radius is a function of true airspeed and bank angle. Rate of turn is a function of true airspeed and turn radius. I'd have to write out and substitute the equations / convert units, but this means bank angle should be be atan(c*turn rate * GS), where c is some constant I need to solve for. Report only if GS is greater than 30 (?) knots.

[...]

PR #439, opened 2016-06-08
Attitude estimation from GPS data

AvSquirrel:
This is an implementation of GPS-derived estimates for pitch and roll angles, as requested on #136 and initially proposed on #340.

The method and performance is similar to the 10 Hz method described in the last few comments on #136.  I've tuned it for performance at 5 Hz, and have performed driving tests on this version using both a BU-353-S4 (SiRF IV) and RY835AI (u-blox 8). Performance on both GPS types felt responsive with minimal lag. In theory, this should also work with a MTK3339 configured with the setup script found in`test/mtk_config.sh`. 

If AHRS hardware is not installed or enabled on the Settings page, this method will automatically run in the background if a GPS is installed and has a position fix. Users must still enable a new "AHRS Output to EFB" option to use the GDL90 AHRS output in WingX, iFly, etc. 

If the "AHRS Output to EFB" option is disabled (either manually, or automatically after ForeFlight detection), attitude data will still  be available over the web interface. This will allow ForeFlight users to view attitude data on the GPS/AHRS UI page, or using 3rd party web tools such as [Pi2D2](https://github.com/Bagarre/Pi2D2).

Other enhancements include:
- Optional recording of GPS attitude performance data to SQLite logs
- AHRS hardware is no longer disabled if ForeFlight is detected; only the GDL90 ouput is suppressed. This applies to both GPS attitude estimate and hardware AHRS connections.
- Tidied up the Status page by removing clock data and moving errors to the bottom of the page (#426)


cyoung:
Nice!

main/network.go
```diff
@@ -496,10 +496,10 @@ func ffMonitor() {
 		}
 		if strings.HasPrefix(s, "i-want-to-play-ffm-udp") || strings.HasPrefix(s, "i-can-play-ffm-udp") || strings.HasPrefix(s, "i-cannot-play-ffm-udp") {
 			p.FFCrippled = true
-			//FIXME: AHRS doesn't need to be disabled globally, just messages need to be filtered.
-			globalSettings.AHRS_Enabled = false
+			//FIXME: AHRS output doesn't need to be disabled globally, just on the ForeFlight client IPs.
+			globalSettings.AHRS_GDL90_Enabled = false
 			if !ff_warned {
-				e := errors.New("Stratux is not supported by your EFB app. Your EFB app is known to regularly make changes that cause compatibility issues with Stratux. See the README for a list of apps that officially support Stratux.")
+				e := errors.New("Stratux AHRS is not currently supported by ForeFlight, and AHRS output has been disabled for all connected clients. See the README for a list of apps that officially support Stratux.")
```
cyoung: Not changing this.

cyoung:
Some thoughts:

* ForeFlight message should stay as-is.
* I'd rather skip adding another option in the UI. Fewer buttons and knobs, the better. This is a setting that can be auto-set. If AHRS is "On" and GPS is "On" and a GPS is available but AHRS hardware not, then use GPS pseudo-attitude and indicate on the main page the AHRS source. If AHRS "On" and AHRS hardware, prefer that.
* `processNMEALine()` is becoming a monster. We should brainstorm on how to break out different GPS parsing into different sub-parts. Setting time, setting `mySituation`, setting pseudo-attitude, and general sanity checks can be broken out.

AvSquirrel:
> ForeFlight message should stay as-is.

I suggested a change because the current message doesn't tell users why the message is being generated, nor does it give any indication that a function of the software is being disabled. I'll revert it, but I disagree with how this is phrased.

> I'd rather skip adding another option in the UI.

In the long-term, we should be able to get away from *any* UI configuration options for AHRS. 

I've been dabbling with non-DMP methods on the 6050 and 9250 (i.e. direct register reads) and it should be possible to auto-detect a limited set of supported devices by scanning the I2C bus for the expected hardware addresses, and pinging the "WHO_AM_I" registers to confirm hardware versions. GPS augmentation (groundspeed and/or turn rate) should be automatically included in the full AHRS solution as needed (or as a standalone estimation for devices without 6-axis / 9-axis hardware).

However, as things stand now, hardware AHRS is still very experimental. If you're going to keep a development feature in a production branch, it needs to be configurable. I've been using the port 80 output (web UI and Pi2D2) for my testing, and there's no need to shut down the whole I2C bus if I just want to suppress the port 4000 AHRS output to make ForeFlight happy.

Maybe an automated per-client toggle in network.go might be a way around that... thinking out loud, is there any reason why we couldn't use `ffmonitor()` to set `outSockets[ip].Capability = NETWORK_GDL90_STANDARD` if ForeFlight is detected, and `outSockets[ip].Capability = NETWORK_GDL90_STANDARD |  NETWORK_AHRS_GDL90` in other cases?

> processNMEALine() is becoming a monster.

Can't disagree. Subfunctions for each NMEA message type would certainly help. Channels could be a cleaner way of sending data to the pseudoattitude function. Maybe a goal for v0.10?

AvSquirrel:
Message is reverted in c5f2193.

skypuppy:
gpsd has already done all that hard work.  <sigh>

cyoung:
>I suggested a change because the current message doesn't tell users why the message is being generated, nor does it give any indication that a function of the software is being disabled.

Actually it does tell the user why. The reason is that Stratux is not supported by ForeFlight and ForeFlight is known to have made changes that have caused compatibility issues and could potentially do so in the future. Is there anything in dispute about this statement? The AHRS "bug" which causes all input to be disabled when Stratux AHRS is detected is being reverted so they say. To say only that "Stratux AHRS" is not supported is unclear and seems to imply that ANY functions of Stratux are supported by ForeFlight. They are not. Stratux software is being put into hardware builds for users that will potentially never upgrade their unit, so making a specific error message which is scantily accurate in the present and surely inaccurate in the future does not make sense.

It's a good idea to warn about disabling the function. Perhaps we should just disable GDL90 AHRS output on a per-client basis instead of globally, and not have to confuse users with AHRS warnings when most people don't even have the hardware.

...Or we could just wait for the change they've said they made and not bother.

AvSquirrel:
What I take issue with is how the message has been presented:

`Stratux is not supported by your EFB app`

There's no dispute for the lack of official support, nor that questions have been raised about future compatibility between Stratux and ForeFlight. The average user should figure out fairly quickly that this refers to ForeFlight, but may be confused about why an error message has appeared if everything seems to be working fine. That requires an understanding of the difference between "support" and "compatibility", but an error message isn't the place for that.

`Your EFB app is known to regularly make changes that cause compatibility issues with Stratux.`

It's fair to point out that there have been changes that broke Stratux compatibility. There was the change in v7.5 that cut off the X-Plane protocol hack we were using to send attitude data, and v7.6 had the messed-up filtering of GDL90 AHRS messages that broke everything. However, I do think it's a stretch to call something that's happened twice "regular."

`See the README for a list of apps that officially support Stratux.`

That part is problematic. A lack of updates to project documentation has pointed users to outdated and/or inaccurate information. As of this afternoon, README.md states:

> Use Pi 2 and ForeFlight 7.3.1 (1792) (Sep 18, 2015).

and

> Tested weather/traffic displays:  ForeFlight 7+ - weather, traffic, AHRS.

Stratux.me also implies that ForeFlight has full support / compatibility:

image

> It's a good idea to warn about disabling the function. Perhaps we should just disable GDL90 AHRS output on a per-client basis...

Written and tested. Say the word and I'll merge this branch [this branch](https://github.com/AvSquirrel/stratux/compare/gpsattitude-june2016...AvSquirrel:disable-ff-ahrs-by-port) into `gpsattitude-june2016`.

Features:
- Per-port disabling of NETWORK_AHRS_GDL90 flag if the "play-ffm" strings are detected
- Per-port re-enabling of NETWORK_AHRS_GDL90 flag if the "play-ffm" strings have not been seen in 20 seconds. This allows app switching on the same device to work normally
- Adds an error message (`AHRS output has been disabled on 192.168.10.xx:4000`) when GDL90 AHRS output is disabled to any given port
- Updated README to account for changes to hardware / software compatibility over the last several months

I flew on it for a couple hours this morning, with GDL90 GPS attitude output enabled. Using two iOS devices, I was able to concurrently run the AHRS display on WingX on one device, while feeding the "standard" GDL90 wx / traffic / position to ForeFlight on the other. I was also able to toggle back and forth between ForeFlight and WingX on the same device, and the AHRS output was automatically enabled and disabled as needed.

jzeevi:
And, what happens if it comes back in the next version of FF? This is a path to a support/development nightmare.

 

I vote for leaving as it is now and see what develops (see what I did there??? J)

cyoung:
You're right, I need to get back on documentation - especially to the ends of app compatibility. There have been some other neat things happening, like WingX's firmware auto-update for Stratux, which need documenting.

main/ry835ai.go
```diff
@@ -436,6 +471,362 @@ func setTrueCourse(groundSpeed uint16, trueCourse float64) {
...
+	} else if length == 1 {
+		//log.Printf("myGPSPerfStats has one data point. Setting statistics to zero.\n")
+		myGPSPerfStats[index].gpsTurnRate = 0
+		myGPSPerfStats[index].gpsPitch = 0
+		myGPSPerfStats[index].gpsRoll = 0
+		return false
+	}
+
+	// check if GPS data was put in the structure more than three seconds ago -- this shouldn't happen unless something is wrong.
+	if (stratuxClock.Milliseconds - myGPSPerfStats[index].stratuxTime) > 3000 {
```
cyoung: Mark.

main/ry835ai.go
```diff
@@ -1438,6 +1923,43 @@ func makeAHRSGDL90Report() {
...
 
+func gpsAttitudeSender() {
+	timer := time.NewTicker(100 * time.Millisecond) // ~10Hz update.
+	for {
+		<-timer.C
+		myGPSPerfStats = make([]gpsPerfStats, 0) // reinitialize statistics on disconnect / reconnect
+		for globalSettings.GPS_Enabled && globalStatus.GPS_connected /*&& globalSettings.GPSAttitude_Enabled*/ && !(globalSettings.AHRS_Enabled) {
+			<-timer.C
+
+			if mySituation.Quality == 0 || !calcGPSAttitude() {
```
cyoung: Mark.

main/ry835ai.go
```diff
@@ -436,6 +471,362 @@ func setTrueCourse(groundSpeed uint16, trueCourse float64) {
...
+		myGPSPerfStats[index].gpsPitch = 0
+		myGPSPerfStats[index].gpsRoll = 0
+		log.Printf("GPS attitude: GPS data is more than three seconds old. Setting attitude to zero.\n")
+		return false
+	}
+
+	// check time interval between samples
+	t1 := myGPSPerfStats[index].nmeaTime
+	t0 := myGPSPerfStats[index-1].nmeaTime
+	dt := t1 - t0
```
cyoung: 3

main/ry835ai.go
```diff
@@ -541,6 +947,7 @@ func processNMEALine(l string) (sentenceUsed bool) {
 			}
 
 			tmpSituation.LastFixSinceMidnightUTC = float32(3600*hr+60*min) + float32(sec)
+			thisGpsPerf.nmeaTime = tmpSituation.LastFixSinceMidnightUTC
```
cyoung: 4

main/ry835ai.go
```diff
@@ -875,6 +1305,9 @@ func processNMEALine(l string) (sentenceUsed bool) {
 		}
 
 		tmpSituation.LastFixSinceMidnightUTC = float32(3600*hr+60*min) + float32(sec)
+		if globalStatus.GPS_detected_type != GPS_TYPE_UBX {
+			thisGpsPerf.nmeaTime = tmpSituation.LastFixSinceMidnightUTC
```
cyoung: 4

main/ry835ai.go
```diff
@@ -969,6 +1427,9 @@ func processNMEALine(l string) (sentenceUsed bool) {
 			return false
 		}
 		tmpSituation.LastFixSinceMidnightUTC = float32(3600*hr+60*min) + float32(sec)
+		if globalStatus.GPS_detected_type != GPS_TYPE_UBX {
+			thisGpsPerf.nmeaTime = tmpSituation.LastFixSinceMidnightUTC
```
cyoung: 4

main/ry835ai.go
```diff
@@ -436,6 +471,362 @@ func setTrueCourse(groundSpeed uint16, trueCourse float64) {
...
+	var valid bool
+	var lengthHeading, lengthSpeed int
+
+	center := float64(myGPSPerfStats[index].nmeaTime) // current time for calculating regression weights
+	halfwidth := float64(1.4)                         // width of regression evaluation window. Default of 1.4 seconds for 5 Hz sampling; can increase to 3.5 sec @ 1 Hz
+
+	// frequency detection
+	tempSpeedTime = make([]float64, 0)
+	for i := 1; i < length; i++ {
+		dt = myGPSPerfStats[i].nmeaTime - myGPSPerfStats[i-1].nmeaTime
```
cyoung: 4

main/ry835ai.go
```diff
@@ -436,6 +471,362 @@ func setTrueCourse(groundSpeed uint16, trueCourse float64) {
...
+			log.Printf("GPS attitude: Something went wrong rebasing the time.\n")
+			return false
+		}
+
+	}
+
+	// If all of the bounds checks pass, begin processing the GPS data.
+
+	// local variables
+	var headingAvg, dh, v_x, v_z, a_c, omega, slope, intercept, dt_avg float64
```
cyoung: 5

main/ry835ai.go
```diff
@@ -436,6 +471,362 @@ func setTrueCourse(groundSpeed uint16, trueCourse float64) {
...
+		// Vertical speed regression estimate.
+		slope, intercept, valid = linRegWeighted(tempSpeedTime, tempVV, tempRegWeights)
+		if !valid {
+			log.Printf("GPS attitude: Error calculating vertical speed regression from UBX position messages")
+			return false
+		} else {
+			v_z = (slope*float64(myGPSPerfStats[index].nmeaTime) + intercept) // units are feet per sec; no conversion needed
+		}
+
+	} else { // If we need to parse standard NMEA messages, determine if it's RMC or GGA, then fill the temporary slices accordingly. Need to pull from multiple message types since GGA doesn't do course or speed; VTG / RMC don't do altitude, etc. Grrr.
```
cyoung: 6

main/ry835ai.go
```diff
@@ -436,6 +471,362 @@ func setTrueCourse(groundSpeed uint16, trueCourse float64) {
...
+				tempSpeedTime = append(tempSpeedTime, float64(myGPSPerfStats[i].nmeaTime))
+				tempRegWeights = append(tempRegWeights, triCubeWeight(center, halfwidth, float64(myGPSPerfStats[i].nmeaTime)))
+			}
+		}
+		lengthSpeed = len(tempSpeed)
+		if lengthSpeed == 0 {
+			log.Printf("GPS Attitude: No groundspeed data could be parsed from NMEA RMC messages\n")
+			return false
+		} else if lengthSpeed == 1 {
+			v_x = tempSpeed[0] * 1.687810
```
cyoung: 7

main/ry835ai.go
```diff
@@ -436,6 +471,362 @@ func setTrueCourse(groundSpeed uint16, trueCourse float64) {
...
+		} else {
+			slope, intercept, valid = linRegWeighted(tempSpeedTime, tempSpeed, tempRegWeights)
+			if !valid {
+				log.Printf("GPS attitude: Error calculating speed regression from NMEA RMC position messages")
+				return false
+			} else {
+				v_x = (slope*float64(myGPSPerfStats[index].nmeaTime) + intercept) * 1.687810 // units are knots, converted to feet/sec
```
cyoung: 7

main/ry835ai.go
```diff
@@ -436,6 +471,362 @@ func setTrueCourse(groundSpeed uint16, trueCourse float64) {
...
+				//log.Printf("Avg VV %f calculated from %d GGA messages\n", v_z, lengthSpeed) // DEBUG
+			}
+		}
+
+	}
+
+	// If we're going too slow for processNMEALine() to give us valid heading data, there's no sense in trying to parse it.
+	// However, we need to return a valid level attitude so we don't get the "red X of death" on our AHRS display.
+	// This will also eliminate most of the nuisance error message from the turn rate calculation.
+	if v_x < 6 { // ~3.55 knots
```
cyoung: 7

main/ry835ai.go
```diff
@@ -436,6 +471,362 @@ func setTrueCourse(groundSpeed uint16, trueCourse float64) {
...
+
+		myGPSPerfStats[index].gpsPitch = 0
+		myGPSPerfStats[index].gpsRoll = 0
+		myGPSPerfStats[index].gpsTurnRate = 0
+		myGPSPerfStats[index].gpsLoadFactor = 1.0
+		mySituation.GPSTurnRate = 0
+
+		// Output format:GPSAtttiude,seconds,nmeaTime,msg_type,GS,Course,Alt,VV,filtered_GS,filtered_course,turn rate,filtered_vv,pitch, roll,load_factor
+		buf := fmt.Sprintf("GPSAttitude,%.1f,%.2f,%s,%0.3f,%0.3f,%0.3f,%0.3f,%0.3f,%0.3f,%0.3f,%0.3f,%0.3f,%0.3f,%0.3f\n", float64(stratuxClock.Milliseconds)/1000, myGPSPerfStats[index].nmeaTime, myGPSPerfStats[index].msgType, myGPSPerfStats[index].gsf, myGPSPerfStats[index].coursef, myGPSPerfStats[index].alt, myGPSPerfStats[index].vv, v_x/1.687810, headingAvg, myGPSPerfStats[index].gpsTurnRate, v_z, myGPSPerfStats[index].gpsPitch, myGPSPerfStats[index].gpsRoll, myGPSPerfStats[index].gpsLoadFactor)
```
cyoung: 8

main/ry835ai.go
```diff
@@ -436,6 +471,362 @@ func setTrueCourse(groundSpeed uint16, trueCourse float64) {
...
+		if myGPSPerfStats[i].coursef >= 0 { // negative values are used to flag invalid / unavailable course
+			tempHdg = append(tempHdg, float64(myGPSPerfStats[i].coursef))
+			tempHdgTime = append(tempHdgTime, float64(myGPSPerfStats[i].nmeaTime))
+		}
+	}
+	//log.Printf("\n")
+	//log.Printf("tempHdg: %v\n", tempHdg)
+
+	// Next, unwrap the heading so we don't mess up the regression by fitting a line across the 0/360 deg discontinutiy
+	lengthHeading = len(tempHdg)
```
cyoung: 9

main/ry835ai.go
```diff
@@ -436,6 +471,362 @@ func setTrueCourse(groundSpeed uint16, trueCourse float64) {
...
+		}
+		return false
+	}
+
+	// Finally, calculate turn rate as the slope of the weighted linear regression of unwrapped heading.
+	slope, intercept, valid = linRegWeighted(tempHdgTime, tempHdgUnwrapped, tempRegWeights)
+
+	if !valid {
+		log.Printf("GPS attitude: Regression error calculating turn rate")
+		return false
```
cyoung: 10

main/ry835ai.go
```diff
@@ -436,6 +471,362 @@ func setTrueCourse(groundSpeed uint16, trueCourse float64) {
...
+				 /__|
+			   X   a_c
+				\_________________ [For the purpose of this comment, " X" is an airplane in a 20 degree bank. Use your imagination, mkay?)
+
+			Resultant acceleration a_r is what the wings feel; a_r/a_z = load factor. Anyway, trig out the bank angle:
+
+				bank angle = atan(a_c/a_z)
+						   = atan(v*omega/g)
+
+				wing loading = sqrt(a_c^2 + a_z^2) / g
```
cyoung: 11

main/ry835ai.go
```diff
@@ -436,6 +471,362 @@ func setTrueCourse(groundSpeed uint16, trueCourse float64) {
...
+
+		*/
+
+		g := 32.174                                        // ft-s^-2
```
cyoung: 7

web/plates/settings.html
```diff
@@ -64,6 +64,12 @@
 		<div class="panel panel-default">
 			<div class="panel-heading">Configuration</div>
 			<div class="panel-body">
+				<div class="form-group">
```
cyoung: 12

cyoung:
Have any demo data? Flown with it? If not I'll give it a test this evening/tomorrow and post a stratux.sqlite.

AvSquirrel:
> Have any demo data? Flown with it?

Here's a brief demo reel. Didn't have sqlite enabled, but I put it through standard rate and steep turns. This used a VK-172 as a position source. At 5 Hz, it's just slightly laggy, but for the most part felt responsive for the types of manuevers one would be doing when one needed an attitude source. This should be more than fast enough to compensate accelerometer inputs for centripetal acceleration.

https://vid.me/AzBV

cyoung:
Moving to https://github.com/cyoung/stratux/tree/gpsattitude, hope to get it packaged for testers.

Issue #298, opened 2016-03-01
Faux tail numbers of "N" and "00000000" with new dump1090


Ergonomicmike:
1. Stratux version: 0.7b3 + 855b "alpha"
2. Stratux config: dual SDR, GPS yes (BU-353), AHRS no; battery, battery cable: Anker E5, Anker cable.
3. EFB app and version (iFly v9.4 RC2). EFB platform (Android KK). EFB hardware (Azus Zenad S 8).
4. Description of your issue. 

I noticed two goofy tail numbers today. "N" and "00000000." (Eight zeros.) Here's the screen shot:
http://postimg.org/image/yx922cu5j/

And here's a clip from es.log
```

{"Icao_addr":11097569,"DF":20,"CA":0,"TypeCode":19,"SubtypeCode":1,"SBS_MsgType":5,"SignalLevel":0.001005,"Tail":"00000000","Squawk":null,"Emitter_category":null,"OnGround":false,"Lat":null,"Lng":null,"Position_valid":false,"NACp":null,"Alt":41000,"AltIsGNSS":false,"GnssDiffFromBaroAlt":null,"Vvel":null,"Speed_valid":false,"Speed":null,"Track":null,"Timestamp":"2016-02-29T17:54:05.797Z"}
49016938747301,
```

Nokomis449:
Both of those "tail numbers" show up in a flightaware search with local and international flights between Europe & the Amaricas, so it's not a bug that's limited to Stratux although it could be a dump1090 bug.  But it's most likely some of your aforementioned chemtrail conspiracy aircraft flown by martians for the Jade Helm exercises.  I'm keeping my tinfoil hat handy just in case.

Ergonomicmike:
Interesting. I wonder how they display at ATC? (peepsnet? Or do we have to have Clearances for you to tell us?) 

If this issue isn't unique to Stratux but an "up the chain" problem, should I close this issue then? It's not like a bizarre call sign is a deal breaker.

peepsnet:
We do not care at all about your ADS-B!!! 
Our displays do not look at adsb data. We associate your squawk code to flight info.

That could be just miscondigured ADS-B eqp. I would like to also see the ICAO HEX code for the targets at the same time. I know you could grep log files but is can be easier then that. 

Ill add another column to the traffic page for that datdata or if @avsquirrel could get that. I know he now has it share with the callsign until the callsign is known but it seems like too many times I need the hex code along with the callsign

[...]

PR #442, opened 2016-06-20
Automatic translation of Mode S codes to US/CA tail numbers

AvSquirrel:
There was some discussion a few months back (#298) about goofy callsigns being transmitted by certain aircraft: "00000000", "N", and everyone's favorite, "BWIFATTY". Consensus was that those callsigns were intentionally entered into the FMS by their flightcrews, and there was some discussion about either filtering "unusual" callsigns, or translating the Mode S code to determine the actual tail number. This is an implementation of the latter.

In both the United States and Canada, civil aircraft registrations (N-number and C-numbers) are related to the 24-bit Mode S code through a 1-to-1 mathematical relationship. In other words, given a specific 24-bit code, it is possible to determine what tail number it belongs to, and vice versa.

The method for calculating US/CAN registrations is implemented in the following way:
- Each time a traffic message is decoded, `parseDownlinkReport()` or `esListen()` checks the `traffic[]` structure to determine if that Mode S code is being tracked. If not, a new traffic object is initialized -- this is current Stratux behavior.
- This change adds a call to a new `icao2reg()` function during traffic object initialization.
- `icao2reg` determines if the Mode S code is in the correct range for US / Canada civil registrations, and if so, decodes it to a registration code. That code is returned as a string.
- Upon successful decoding, the registration code is assigned to both `ti.Tail` (callsign) and a new `ti.Reg` variable (registration).
- Subsequent ADS-B traffic messages will typically provide callsign information; at that point, `ti.Tail` will be overwritten with whichever callsign the aircraft is broadcasting, and will be used for the GDL90 output. However, `ti.Reg` will retain the registration code for that aircraft.
- Traffic UI by default will still show callsign (`ti.Tail`). A toggle has been added to allow the decoded registration (`ti.Reg`) to be shown.
- "Show Distance" toggle has also been cleaned up to remove the "GPS Connected" indicator. If GPS disconnects, the text will change to "Show Distance N/A"  and will be grayed out. Addresses #426.

image

image

### DIFF
--- a/main/traffic.go
+++ b/main/traffic.go
@@ -13,6 +13,7 @@ import (
 	"bufio"
 	"encoding/hex"
 	"encoding/json"
+	"fmt"
 	"log"
 	"math"
 	"net"
@@ -75,7 +76,8 @@ const (
 
 type TrafficInfo struct {
 	Icao_addr           uint32
-	Tail                string
+	Reg                 string    // Registration. Calculated from Icao_addr for civil aircraft of US registry.
+	Tail                string    // Callsign. Transmitted by aircraft.
 	Emitter_category    uint8     // Formatted using GDL90 standard, e.g. in a Mode ES report, A7 becomes 0x07, B0 becomes 0x08, etc.
 	OnGround            bool      // Air-ground status. On-ground is "true".
 	Addr_type           uint8     // UAT address qualifier. Used by GDL90 format, so translations for ES TIS-B/ADS-R are needed.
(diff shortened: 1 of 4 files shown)


jzeevi:
Clever. Is the translation tail number independent? For example, If I buy a plane and later change the tail number, does my Mode S code change? Or is there a database that needs to be downloaded periodically?

cyoung:
>Clever. Is the translation tail number independent? For example, If I buy a plane and later change the tail number, does my Mode S code change? Or is there a database that needs to be downloaded periodically?

At least for the US, all possible tail numbers are already accounted for in their algorithm. You'd change the Mode S code on your transponder if your tail number changed.

AvSquirrel:
> Clever. Is the translation tail number independent? 

 > At least for the US, all possible tail numbers are already accounted for in their algorithm.

Right. There's no lookup table for this method; it's purely algorithmic. Changing your tail number in either country changes the Mode S code. The avionics manufacturers can actually use this to their advantage: During installation of Mode S tranponders from Trig, Garmin, Dynon, etc., if you enter an N-number as the tail number, they will  automatically calculate the Mode S 24-bit code.

Canada is especially easy, since registration codes always use the same format (CFxxx, CGxxx, or CIxxx, where 'x' is a letter A-Z). Therefore, the Mode S code is just a base-26 expression of the tail number, after adding the 0xC00001 offset. The last three letters map as A=0, B=1, ... Z = 25. The second letter maps as F = 0, G = 1, I = 2.

FAA registrations are more convoluted because of the variable length of tail number and the rules surrounding where letters / numbers can be used.

PR #444, opened 2016-06-21
GPIO fan control script

egid:
This adds /u/helno's GPIO fan control script. It doesn't run automatically - it would need an edit to crontab, and I don't want to hook up anything that's aimed at a small set of users.

That said, it seems like the kind of utility that wouldn't hurt to distribute with the main release, even just to allow other folks to hack on it with their own hardware.


cyoung:
@egid's inaugural PR ?

PR #461, opened 2016-07-15
GPS accuracy on Status page

egid:
This is a really simple tweak based on something I noticed in my testing use case.

I've been doing some device testing, and it's useful to know the accuracy of the GPS without tapping around and losing the ability to see messages/min for UAT/1090, etc.

This adds `mySituation.Accuracy` to the output of `updateStatus`, and handles it on the front end. To keep from linewrapping on small screens, I simplified `WAAS / EGNOS` to just `WAAS`, although I could see this being less than good. If it's requested, I can probably fiddle with the layout of that component and create more space on the values side; there's room to take away from the labels.

I'm new to playing around in the angular side - let me know if I'm doing Bad Things by manipulating this into a String or something!

screenshot 2016-07-15 00 49 19

### DIFF
--- a/main/gen_gdl90.go
+++ b/main/gen_gdl90.go
@@ -707,7 +707,7 @@ func cpuTempMonitor() {
 
 func updateStatus() {
 	if mySituation.Quality == 2 {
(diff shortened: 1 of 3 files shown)


cyoung:
Looks fine, thanks.

Issue #400, opened 2016-04-24
GPS not detected when plugged in AFTER Stratux start


cyoung:
1. Stratux version: v0.8r2 (a2ca212a52)
2. Stratux config:

    SDR
      [x] single
      [ ] dual

    GPS
      [x] yes
      [ ] no
      type:

    AHRS
      [ ] yes
      [x] no

    power source: CanaKit 2.5A wall wart

    usb cable: built-in

3. EFB app and version: None

    EFB platform: N/A

    EFB hardware: N/A

4. Description of your issue:

When GPS is plugged in after Stratux start, GPS status stays in "disconnected" state (per WebUI).

GPS comes up on `/dev/ttyACM0`. Pi3.

AvSquirrel:
Can't reproduce on Pi2 / Wheezy or Pi3 / Jessie (with or without `pi3-miniuart-bt` overlay)

Is anything attached as `/dev/ttyUSB0`?

skypuppy:
I have seen this lately, too, with a USB GPS and with the serial GPS.  0.8r2 with image downloaded from stratux.me.  I don't recall it occurring when I loaded the full blown jessie, then sh'ed 8.8r1 in.

cyoung:
@AvSquirrel - there was. It seems like it does it mostly when removing one GPS device and inserting another. Probably has to do something with some baud rate hysteresis.

AvSquirrel:
> Is anything attached as `/dev/ttyUSB0`?
> there was.

That would do it. This goes back to the BU-353 support (#231) that was committed a few months ago.

If `initGPSSerial()` [sees `/dev/ttyUSB0`, it assumes that there is a BU-353 attached](https://github.com/cyoung/stratux/blob/master/main/ry835ai.go#L128-L137) and starts configuration on that port, ignoring any devices on `/dev/ttyACM0` or `/dev/ttyAMA0`.

I can address this when I rework the GPS initialization. Moving `/dev/ttyACM0` ahead of `/dev/ttyUSB0` in priority order should resolve this for the recommended use cases (VK-172 or other u-blox over USB). Not sure yet how I want to handle GPIO.

peepsnet:
Could we do some kind of detect all GPS types and allow user to select from a list if more then one??

[...]

PR #464, opened 2016-07-16
Update setup scripts to disable console on ttyAMA0

AvSquirrel:
See #400.

This updates two setup scripts (mkimg.sh and update_footer.sh) to look for and remove any `console=ttyAMA0` references in `/boot/cmdline.txt`, resolving a serial connection issue that caused USB receivers to be unable to connect if installed after boot, and for intermittent disconnects of GPS receivers connected over serial GPIO. 

This also removes an outdated reference to `/etc/inittab` in mkimg.sh.

### DIFF
--- a/image/mkimg.sh
+++ b/image/mkimg.sh
@@ -113,7 +113,6 @@ make
 make install
 
 #disable serial console
-sed -i /etc/inittab -e "s|^.*:.*:respawn:.*ttyAMA0|#&|"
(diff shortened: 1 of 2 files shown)


skypuppy:
Outstanding, AvSquirrel!  This allowed my serial GPS to connect almost immediately instead of the former 8 to 10 minutes.  Woohoo!

PR #465, opened 2016-07-19
Gdl90report

westphae:
This PR is for upgrading the GDL90Report to iLevil's report Package ID Version 1 (current version implemented in stratux is 0).  This adds a few more fields to the report: airspeed, pressure altitude and vertical speed.  Version 1 is the currently supported version in DroidEFB (version 0 wouldn't work; an upgrade was just released today that works with version 1).

This PR is in preparation for AHRS support, so it also adds Slip/Skid, Turn Rate, and GLoad to the stratux Situation struct so they will be automatically sent.

Attitude output tested with DroidEFB and Naviator.  I was unable to find an app that would report pressure altitude from the GDL90Report stream, and I don't have an Apple device to test, so would appreciate if someone with an iPad could test before merging (test both altitude/rate of climb reception and correct communication with Foreflight etc.).  Output looks correct when viewed with netcat, however.

This is my first PR, so I thought I would keep it limited and simple.  Thanks!

### DIFF
--- a/main/network.go
+++ b/main/network.go
@@ -135,7 +135,11 @@ func sendToAllConnectedClients(msg networkMessage) {
 			if sleepFlag {
 				continue
 			}
-			netconn.Conn.Write(msg.msg) // Write immediately.
+			_, err := netconn.Conn.Write(msg.msg) // Write immediately.
+			if err != nil {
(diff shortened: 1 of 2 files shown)


cyoung:
@westphae - thanks for this! Anyone able to test on the iPad apps?

PR #479, opened 2016-08-21
Fixed typo in aliases for setSerial1

tc0nn:
The image has a typo in the aliases for the setSerial1 alias.


peepsnet:
thanks... forgot that one

PR #481, opened 2016-08-23
Use ReplayLog data to create KML files of both traffic and ownship

alexberryman:
Changes to the Settings page allows the user to dump the ReplayLog (traffic+ownship) SQLite file or in the new "GPS Track File" panel to just dump the ownship  `mySituation` GPS data. The new 'Write GPS to file' setting only writes the `mySituation` portion of ReplayLog to limit the growth of the stratux.sqlite file size.

The ReplayLog KML file can be created with animations based on time, or altitude. The altitude filter is useful to remove cruise traffic and view traffic below. Demo: https://zippy.gfycat.com/OfficialFirstChinesecrocodilelizard.webm

The KML files have folders based on the following criteria:
- ownship
- ADS-B Traffic w/min alt < FL180
- ADS-B Traffic w/min alt > FL180
- UAT Traffic (`TARGET_TYPE>1`)

Clicking on traffic trails will display minimum and maximum altitude and provide links to Flightaware.

The only thing I think I am missing is a way to clear the SQLite file from the web UI to either free up space or when starting a new flight.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ all:
 
 xgen_gdl90:
 	go get -t -d -v ./main ./test ./linux-mpu9150/mpu ./godump978 ./mpu6050 ./uatparse
-	go build $(BUILDINFO) -p 4 main/gen_gdl90.go main/traffic.go main/ry835ai.go main/network.go main/managementinterface.go main/sdr.go main/uibroadcast.go main/monotonic.go main/datalog.go main/equations.go
+	go build $(BUILDINFO) -p 4 main/gen_gdl90.go main/traffic.go main/ry835ai.go main/network.go main/export_datalog.go main/managementinterface.go main/sdr.go main/uibroadcast.go main/monotonic.go main/datalog.go main/equations.go
 
 xdump1090:
 	git submodule update --init
--- a/main/datalog.go
+++ b/main/datalog.go
@@ -532,7 +532,7 @@ func isDataLogReady() bool {
 }
 
 func logSituation() {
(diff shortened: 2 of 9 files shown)


peepsnet:
WOW! Nice

cyoung:
Looks pretty good. The biggest issue I have with this is that it opens and runs from the gen_gdl90 daemon. File locking issues.

Breaking out some of the replay log features is a good idea.

How are most people handling .kml files?

Bearcode:
Does the [Write-Ahead Logging](https://sqlite.org/wal.html) in SQLite3 prevent the file lock issues you are concerned with? In my testing I haven't been able to create a file lock error. It looks like there is no issue with concurrent writes from `datalog.go` and reads from `export-datalog.go`.

Here is a [Gist of lsof output](https://gist.github.com/Bearcode/932a0f4f6c1e2a309b5e80ec53543f88) showing multiple locks on the SQLite, `-wal` and `-shm` files while testing concurrent reads and writes.

Is there any part of the code base that doesn't run under gen_gdl90 that I can look at as an example for breaking out these functions?

evannadeau:
Any chance there can be future movement on this?  How about saving the files to another partition on the microSD?  I have a 32GB one that I put Stratux on.  It is still however only using the standard 2GB partition.  Another 28GB (after formatting) still kicking around I'm sure.

alexberryman:
Stale PR

benlawraus:
Hi, can the work of @alexberryman be merged please ?  In developer mode maybe?  This would make stratux the only adsb device with flight logging....

Issue #483, opened 2016-08-27
Serial GPS is not properly detected on RPi 3


jamez70:
If using a serial gps connected to the UART pins of the RPi 3, the software defaults to /dev/ttyAMA0 which is used by bluetooth. The GPS will be present on /dev/ttyS0

cyoung:
Fix currently implemented, see PR #484.

PR #394, opened 2016-04-20
GPS UART fix for RPi 3

AvSquirrel:
- Resolves #393 by removing `/dev/ttyS0` from GPS initialization sequence. On the RPi 3, this port does *not* connect to GPIO pins 8 and 10; receivers connected to these pins are therefore not detected.
- Add a interactive script for configuring MTK3339 GPS receivers from the command line

PR #484, opened 2016-08-28
Added fix for serial gps on rPI3

jamez70:
Per issue #483 - Serial GPS was defaulting to /dev/ttyAMA0 instead of /dev/ttyS0 on the RPi 3


cyoung:
See PR #394. Different approach to the same fix, but we've implemented that one already. I think you might be having issues since you're using a different `/boot/config.txt` from the stock image?

PR #486, opened 2016-09-05
RFC: gpsd evaluation

foonix:
This is a basic hack to pull GPS information from gpsd.  There seemed to be some interest in assessing the usefulness of gpsd, so I created this as an excuse to learn Go and play with gpsd.  This is my first Go project so any feedback is appreciated.  If anyone can think of any features that might help with assessment, please let me know.

To run this you will need to connect your Pi to the internet, `apt-get install gpsd gpsd-clients`, and enable/start the service.

Some initial observations:
- GPS hotplug works great.  Basically a non-issue on the stratux level.
- It will take data from more than one GPS, however, there is not (currently) any code for selecting which report to use.  Whatever is the latest to be received is what is used.
- The gpsd shipping with Raspbian Jessie (3.11) does not have udev rules for ublox > 6.  My ublox7 receiver worked fine after adding rules.  However, it may be a good idea to install a newer version of gpsd, which is available in Stretch (3.16).
- This code does not coexist with native ry835ai.go, so most of that is disabled.  If there is some interest I would be willing to add it.
- stratux has no notion of a 2D fix that I can see.  So 2D fix is presented as 3D at the last known altitude.  It could also ignore 2D fixes.  Is there a preference here?
- I can't find a reliable and easy way to tell if the fix is WAAS or not.  So, all fixes are presented as a normal 3D fix for now.  See: http://catb.org/gpsd/faq.html#waas
- There is an ATT status gpsd provides which could probably be used for AHRS.  However, the only two products I could confirm support it seem quite expensive. This might be worth further investigation.  Gpsd its self doesn't seem to do AHRS calculations but it looks for proprietary NMEA messages from devices that do them.  I added handling for these messages, but can't test it.

For troubleshooting I recommend to try the `cgps` gps command line client to confirm gpsd is working and receiving a fix.  Stratux will connect to gpsd on startup.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ all:
 
 xgen_gdl90:
 	go get -t -d -v ./main ./test ./linux-mpu9150/mpu ./godump978 ./mpu6050 ./uatparse
-	go build $(BUILDINFO) -p 4 main/gen_gdl90.go main/traffic.go main/ry835ai.go main/network.go main/managementinterface.go main/sdr.go main/ping.go main/uibroadcast.go main/monotonic.go main/datalog.go main/equations.go
+	go build $(BUILDINFO) -p 4 main/gen_gdl90.go main/traffic.go main/ry835ai.go main/network.go main/managementinterface.go main/sdr.go main/ping.go main/uibroadcast.go main/monotonic.go main/datalog.go main/equations.go main/gpsd.go
 
 xdump1090:
 	git submodule update --init
--- a/image/10-stratux.rules
+++ b/image/10-stratux.rules
@@ -6,9 +6,9 @@
 #   ublox7: VK-172, RY725AI
 #   ublox6: VK-162
 
-SUBSYSTEMS=="usb", ATTRS{idVendor}=="1546", ATTRS{idProduct}=="01a8", SYMLINK+="ublox8"
-SUBSYSTEMS=="usb", ATTRS{idVendor}=="1546", ATTRS{idProduct}=="01a7", SYMLINK+="ublox7"
-SUBSYSTEMS=="usb", ATTRS{idVendor}=="1546", ATTRS{idProduct}=="01a6", SYMLINK+="ublox6"
+SUBSYSTEMS=="usb", ATTRS{idVendor}=="1546", ATTRS{idProduct}=="01a8", SYMLINK+="ublox8", SYMLINK+="gps%n", TAG+="systemd", ENV{SYSTEMD_WANTS}="gpsdctl@%k.service"
+SUBSYSTEMS=="usb", ATTRS{idVendor}=="1546", ATTRS{idProduct}=="01a7", SYMLINK+="ublox7", SYMLINK+="gps%n", TAG+="systemd", ENV{SYSTEMD_WANTS}="gpsdctl@%k.service"
+SUBSYSTEMS=="usb", ATTRS{idVendor}=="1546", ATTRS{idProduct}=="01a6", SYMLINK+="ublox6", SYMLINK+="gps%n", TAG+="systemd", ENV{SYSTEMD_WANTS}="gpsdctl@%k.service"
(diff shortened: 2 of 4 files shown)


cyoung:
@foonix - Thank you for submitting this. Just looking at it now, sorry it took so long.

>stratux has no notion of a 2D fix that I can see. So 2D fix is presented as 3D at the last known altitude. It could also ignore 2D fixes. Is there a preference here?

EFBs take the 3D position and calculate terrain and/or traffic alerts based on this data. Most EFBs don't account for an invalid altitude input. If we fudge the 3D fix by using last known altitude, that has potential to cause some ugly bugs. 2D-only fixes are mostly edge cases anyways, if the user is only getting a 2D fix I'd rather they get no information at all, as they should be forced to figure out the problem with their receiver. For that reason, as you noted, most of the stratux code counts on all GPS fixes to be 3D fixes

>I can't find a reliable and easy way to tell if the fix is WAAS or not. So, all fixes are presented as a normal 3D fix for now. See: http://catb.org/gpsd/faq.html#waas

This is where we start asking questions on why we want to use gpsd. We can make the simplifying assumption that all users will be using a u-blox 6/7/8, or we can begin to account for the issues as in your link.
[This](https://github.com/cyoung/stratux/blob/master/main/ry835ai.go#L1224-L1240) is working well for us currently.

cyoung:
I see now that you've implemented the same SBAS 'detection' in `getNMEAName()`. Great, this is a really straightforward implementation looks pretty good to me.

I'm modifying this PR now to merge into https://github.com/cyoung/stratux/tree/gpsd, we can start there.

What remains:

1. gpsd install on images.
2. udev rules that you reference? Which are those?
3. Some update vector, would ideal to send this out via .sh files if possible versus a full image update.
4. Going through `processNMEALine()` one by one and checking each function there. We're doing custom configuration on u-blox receivers and it would be good to check what we need there and what we don't. Some unimplemented features in this code currently are setting system clock, update rate/baud rate changes (if needed).
5. It would be a good time to also overhaul the `isGPSConnected()`, `isGPSValid()`, `isGPSGroundTrackValid()`, and `isGPSClockValid()` functions.

foonix:
Good point on the altitude.  I'll send another PR with the fix.  

`getNMEAName()` is mostly stolen from `processNMEALine()`. :)   I'll take a stab at checking the IDs marked as in solution.  If all else fails we can still snoop NMEA via gpsd.  If stratux only says a fix is WAAS if it can positively identify it is, I think it's fair that people using receivers not on the recommended list might not get their fix to show as WAAS even if it is.


2. The changes in `image/10-stratux.rules` mimic the rules in question.  The original rules are in `/lib/udev/rules.d/60-gpsd.rules` and are (gpsd 3.11):
 ```
# u-blox AG, u-blox 5 (tested with Navilock NL-402U) [linux module: cdc_acm]
ATTRS{idVendor}=="1546", ATTRS{idProduct}=="01a5", SYMLINK+="gps%n", TAG+="systemd", ENV{SYSTEMD_WANTS}="gpsdctl@%k.service"
# u-blox AG, u-blox 6 (tested with GNSS Evaluation Kit TCXO) [linux module: cdc_acm]
ATTRS{idVendor}=="1546", ATTRS{idProduct}=="01a6", SYMLINK+="gps%n", TAG+="systemd", ENV{SYSTEMD_WANTS}="gpsdctl@%k.service"
```
3. The gpsd .deb and dependencies would have to be bundled with the .sh unless we can require an internet connection.  It might be mandatory if we need a newer gpsd version than what is in the repo.
4.   Gpsd has code for negotiating baud rates.  There is a mechanism where gpsd can [communicated with ntpd over shared memory](http://www.catb.org/gpsd/gpsd-time-service-howto.html)
5. To add, I noticed there could be some latency between the gps fix and the ownship report caused by the heartbeat timing loop.  Maybe `makeOwnshipReport()` could read fixes from a channel or be triggered by callback.

PR #491, opened 2016-09-26
Added capability for a custom SSID. 

jamez70:
If the file /etc/hostapd/ssid.conf exists, it uses the contents of this file for the SSID instead of stratux

This allows the updating of the software and allowing the use of a custom SSID

### DIFF
--- a/image/hostapd-edimax.conf
+++ b/image/hostapd-edimax.conf
@@ -1,6 +1,5 @@
 interface=wlan0
 driver=rtl871xdrv
-ssid=stratux
 hw_mode=g
 channel=1
 wme_enabled=1
--- a/image/hostapd.conf
+++ b/image/hostapd.conf
@@ -1,5 +1,4 @@
 interface=wlan0
-ssid=stratux
 hw_mode=g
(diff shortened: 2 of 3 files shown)


peepsnet:
Breaks!

cyoung:
This probably will break with RO implementations.

Started researching this and found `wpa_ctrl` interface for hostapd/wpa_supplicant.

Second example can be used for setting Stratux in client mode and having a nice method to control this via the web interface: http://vlkan.com/wapi/group__hostapd.html.

http://w1.fi/hostapd/devel/wpa__ctrl_8c_source.html

@ssokol - is your RO script run before or after runlevel S?

If after, can merge with a couple changes.

peepsnet:
I improved on this with https://github.com/cyoung/stratux/pull/529

This pull is obsolete

peepsnet:
I believe this could be closed with the implementation of my code

https://github.com/cyoung/stratux/pull/529
and 
https://github.com/cyoung/stratux/pull/550

cyoung:
Thanks.

PR #492, opened 2016-09-28
Added ability to have static IP hosts

jamez70:
This adds the ability to have static IP hosts. I frequently use the stratux with an ethernet adapter, thus having this feature will really speed up development for me, and also allows to have a stratux to be connected on a wired network. The file /etc/stratux-static-hosts.conf would contain lines as such:

192.168.0.2 host1
192.168.0.3 host2

and this would then be appended to the dhcp client list.


cyoung:
This will help a lot for dev.

PR #504, opened 2016-10-03
Developer mode flag handling

jamez70:
Handles a developer mode, with a reload function in the developer menu

Also can enter development mode by clicking the version 7 times in 3 seconds :)


cyoung:
Things we should start moving to developer mode:

1. "PPM Correction" input field.
2. All "Diagnostics" in Settings.
3. "Logs" tab.

PR #505, opened 2016-10-05
Trim leading garbage before dollar sign

jamez70:
Trim leading garbage on GPS received string which would previously return an error. The string is still good other than the leading garbage

### DIFF
--- a/main/ry835ai.go
+++ b/main/ry835ai.go
@@ -394,12 +394,14 @@ func initGPSSerial() bool {
 
 func validateNMEAChecksum(s string) (string, bool) {
 	//validate format. NMEA sentences start with "$" and end in "*xx" where xx is the XOR value of all bytes between
-	if !(strings.HasPrefix(s, "$") && strings.Contains(s, "*")) {
+	if !(strings.Contains(s, "$") && strings.Contains(s, "*")) {
 		return "Invalid NMEA message", false
 	}
-
+	// Trim the leading garbage from the string
+	ds_index := strings.Index(s, "$")
+	s_new := s[ds_index:len(s)]
 	// strip leading "$" and split at "*"
-	s_split := strings.Split(strings.TrimPrefix(s, "$"), "*")
+	s_split := strings.Split(strings.TrimPrefix(s_new, "$"), "*")
 	s_out := s_split[0]
 	s_cs := s_split[1]
 


cyoung:
What's the source of the garbage I wonder?

jamez70:
Thats a great question. It appears to be binary data but not consistent

skypuppy:
partial bit strings, previously ignored.

cyoung:
Well, @jamez70 pasted some output from his GPS (re-paste for posterity if you want) - every line seemed to be valid but with prepended invalid data.

jamez70:
`2016/10/05 03:04:25 GPS error. Invalid NMEA string: Invalid NMEA message
String: �!�%�1$PUBX,00,030425.40,4233.59655,N,08751.83185,W,174.532,D3,4.4,6.9,1.177,58.41,0.502,,0.93,1.31,0.97,12,0,0*77
2016/10/05 03:04:29 GPS error. Invalid NMEA string: Invalid NMEA message
String: ȱ!%!!$PUBX,00,030429.40,4233.59668,N,08751.83167,W,172.755,D3,4.6,7.0,0.793,58.41,0.011,,1.58,2.45,1.87,5,0,0*4C`

I was receiving that garbage in the string

skypuppy:
If I  had access to a logic analyzer again...  Or a bus pirate.  Usually,  it's the remainder of the previous string coming from the GPS *before* code starts reading that bus.  Once, on a chinese gps, it was spurious noise from the gps circuitry but that was only one single chip.  Upon powerup, gps will start spewing strings but code doesn't start looking for it until later and with these boards, there is no buffering prior to port init.

skypuppy:
That looks exactly like that one chip that we tossed.  At times, on the very first string, there is a little bit-twiddling until the speeds start sync-ing..  When it repeats after first valid string, it appears to be a faulty chip.  Hmm, could be RF noise, as well, but that would show up inside the valid strings as well.

skypuppy:
I have a bus pirate on the way.

jamez70:
I have one, a few scopes and some other junk around here

skypuppy:
a logic analyzer is the ideal tool for this, esp if it can convert to ASCII.

skypuppy:
A scope can show if it's RF interference, by voltages being too high or too low compared to standard serial or USB voltages, also with random pulse widths.

skypuppy:
You need to pull out the old digital storage scope.  :)

skypuppy:
Another problem with using a scope is the variable, valid signal widths.

skypuppy:
And I have just exhausted my knowledge of scopes, even if I have 5.  Only one is a digital storage one and it's a small portable unit.

jamez70:
I'll use the 12GHz scope at work to make sure PCIe isn't on the bus

skypuppy:
PCIe on a Pi?

jamez70:
I'm just kidding

skypuppy:
As I was lying down I remembered a much more common occurrence: that of setting the baud rate and/or gps report rate higher than the system can keep up with in a given circumstance.

jamez70:
Well I ran it for 12+ hours now - not one NMEA error

skypuppy:
are you near 4 towers?

jamez70:
What do you mean

peepsnet:
AAARRRRRRR!!!!

jamez70:
14494717_1435593426456722_2492417015899665113_n

cyoung:
@jamez70 https://frillip.com/raspberry-pi-3-uart-baud-rate-workaround/

skypuppy:
@jamez70, yeah, I had read through that before plus it's link, but by post 7 or 9 of it's link, I was completely baffled and had lost my ability to parse English by the last post.

jamez70:
haha

cyoung:
Closing this, I think it's an RPi serial clock issue.

PR #520, opened 2016-11-14
Added seconds tag to stratux.appcache for page reloading if built cod…

jamez70:
…e has changed


cyoung:
Why wasn't `stratuxBuild` enough? It should change every time the web files change.

PR #521, opened 2016-11-14
Possible solution for upper level winds forecast time

jamez70:
The WINDS forecast will show 'from now' instead of '29 days ago'. It may not be the right solution, but at least this is a good placeholder for a change


cyoung:
What about no time calculation at all if greater than 2 days? The display is meant to only show a UAT stream, so all WINDS data should be fresh.

jamez70:
I hate unfresh winds

jamez70:
I'm not exactly sure what you mean. Let me know if you want me to change it, or feel free to change it to what you like.

PR #522, opened 2016-11-14
A font you can actually read on a tablet/phone

jamez70:
I changed the font weight for 'body' because on a tablet/phone the text was hard to read on the web status page

### DIFF
--- a/web/css/main.css
+++ b/web/css/main.css
@@ -237,7 +237,7 @@
 
 
 /* ***************************************************************************
-everything below this comment represents tweeks to the mobile-angular-uis CSS
+everything below this comment represents tweaks to the mobile-angular-uis CSS
 *************************************************************************** */
 
 @font-face {
@@ -396,6 +396,9 @@ pre {
 	border-radius: 0px;
 }
 
+body {
+    font-weight: 400;
+}
 
 /* change right sidebar behavior to always push */
 


jamez70:
Before
screenshot_20161114-000131

jamez70:
After
screenshot_20161114-000141

jamez70:
After:
screenshot_20161114-000808

Ergonomicmike:
For informational purposes (for anyone reading this who interested in CSS in Stratux), a font-weight of "400" is considered "normal." So, in theory, this commit shouldn't have changed anything. But it did.

I'm guessing that what had happened here is that the text that makes up the UAT Statistics inherited a weight less than 400 from somewhere else. (Perhaps from the surrounding div (or divs)?) So by declaring the font-weight for body text to be 400, that resets (and overrides) any other (inherited) setting.

There's nothing wrong with declaring style for body text, and I tend to do it to avoid any unexpected problems down the road. So this is a fine fix.

An additional, perhaps more "elegant," option would also be for someone to go back into the code and create a class selector for the UAT message text that sets the weight to normal, so that that text is unique.

However, that's a lot of work simply to be elegant. Since a font-weight of 400 is "normal," if there are any other instances of text in the html that haven't been styled, they will now be styled to a normal font weight. So there shouldn't be any anomalies caused by this clever commit by Jamez70.

Ergonomicmike:
In fact, looking at a web page of mine, it used to be best practice in the old days to declare the styles for "body," because different browser had different defaults. (So maybe that's the problem this commit fixed?) And, due to a bug in some browser, it used to be recommended that one set the font-size yo 100.01%.

PR #523, opened 2016-11-16
Keep favorite weather stations

jamez70:
Remembers weather stations in the favorite list which does not clear when leaving the weather page

### DIFF
--- a/main/gen_gdl90.go
+++ b/main/gen_gdl90.go
@@ -755,8 +755,8 @@ func registerADSBTextMessageReceived(msg string) {
 	}
 
 	var wm WeatherMessage
-    var dataStr string
-    
+	var dataStr string
+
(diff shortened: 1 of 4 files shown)


cyoung:
At this point I've only looked through the Go portion, but the `[]watchedStationType` could have some thread concurrency issues.

Could we change the `/weatherwatched` ws to work more like the `/traffic` ws, where the current traffic is sent then only changes are sent (versus re-sending all the watched weather reports every 5 seconds)?

jamez70:
Possibly. I had originally had it only send if there was a change, but then when you navigate to the web page you will have no data. Then, I attempted to send when the websocket was connected, then only after a change and that did not work very well either. The problem is if you leave and reopen the web page it does not have any data and would need all of the data in order to display it.

PR #525, opened 2016-11-17
Church up weather status page

jamez70:
Little better appearance tweaks - you may or may not like them

### DIFF
--- a/web/css/main.css
+++ b/web/css/main.css
@@ -131,6 +131,21 @@
 	content: "\f1d9";
 }
 
+.report_TAF {
+	background-color: cornsilk;
+	color: black;
+}
+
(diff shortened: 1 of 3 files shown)


jamez70:
image

web/css/main.css
```diff
@@ -131,22 +131,45 @@
...
+}
+
 .flight_condition_VFR {
+    border-radius: 5px;
 	background-color: forestgreen;
 	color: white;
 }
 
 .flight_condition_MVFR {
+    left-border: 120px;
```
jamez70: Oops. Get rid of that. It has no effect. I was trying to make the oval a bit larger and that is a leftover

PR #526, opened 2016-11-17
GDL90 JSON interface stratux changes

jamez70:
This encapsulates GDL90 traffic over a websocket for broadcasting to remote hosts. The remote host can connect to a stratux over the internet and receive the GDL90 to be rebroadcasted on its local network over UDP. This will be used for remote developers so they can use a remote stratux for software development of Android or iPad applications using real data.

### DIFF
--- a/main/managementinterface.go
+++ b/main/managementinterface.go
@@ -35,6 +35,25 @@ type SettingMessage struct {
 // Weather updates channel.
 var weatherUpdate *uibroadcaster
 var trafficUpdate *uibroadcaster
+var gdl90Update *uibroadcaster
+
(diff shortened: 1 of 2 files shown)


main/managementinterface.go
```diff
@@ -35,6 +35,25 @@ type SettingMessage struct {
 // Weather updates channel.
 var weatherUpdate *uibroadcaster
 var trafficUpdate *uibroadcaster
+var gdl90Update *uibroadcaster
+
+func handleGdl90WS(conn *websocket.Conn) {
```
cyoung: Caps.

PR #529, opened 2016-11-19
Move user configurable options out of hostapd.conf and hostapd-edimax.conf to hostapd.user and build temp hoatapd.conf on the fly with stratux-wifi.sh

peepsnet:
I have changed the code to use hostapd.user, which contains all the user configurable values and modified stratux-pre-start.sh and stratux-wifi.sh to create /tmp/hostapd.conf on the fly by merging the standard /etc/hostapd/hostapd.conf(or hostapd-edimax.conf if pi2) and /etc/hostapd/hostapd.user to build the /tmp/hostpad.conf.

This allows us to overwrite the /etc/hostapd.conf(or hostapd-edimax.conf) whenever we need and keep any user values.

The changes also will create the hostapd.user file if it does not exist and move all the user configurable settings into it. 

I also changed my hostapd_manager.sh to work on hostapd.user now

### DIFF
--- a/__root__stratux-pre-start.sh
+++ b/__root__stratux-pre-start.sh
@@ -15,3 +15,39 @@ if [ -e /root/update*stratux*v*.sh ] ; then
 	fi
 fi
 
+##### Script for setting up new file structure for hostapd settings
+##### Look for hostapd.user and if found do nothing.
+##### If not assume because of previous version and convert to new file structure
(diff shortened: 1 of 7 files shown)


peepsnet:
waiting for one more piece of script!!

peepsnet:
@cyoung ::whistling and tapping toe in the corner::

PR #539, opened 2016-12-04
Gpsd connection management

foonix:
* Monitor gpsd connection failures and automatically reconnect
* Add stratux.conf setting to use a different gpsd server.  Advanced users can use a gpsd on another machine.
* User controls connection from management interface using GPS on/off toggle.  If gpsd has no clients, it will disconnect from the device.  Supposedly this can save power on some devices.
* Log individual gpsd messages only in verbose mode.

The requires the latest version of [stratux/go-gpsd](https://github.com/stratux/go-gpsd)

### DIFF
--- a/main/gen_gdl90.go
+++ b/main/gen_gdl90.go
@@ -984,6 +984,7 @@ type settings struct {
 	ES_Enabled           bool
 	Ping_Enabled         bool
 	GPS_Enabled          bool
+	GpsdAddress          string
 	NetworkOutputs       []networkConnection
 	SerialOutputs        map[string]serialConnection
 	AHRS_Enabled         bool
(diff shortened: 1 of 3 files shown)


main/gen_gdl90.go
```diff
@@ -1046,6 +1047,7 @@ func defaultSettings() {
 	globalSettings.UAT_Enabled = true
 	globalSettings.ES_Enabled = true
 	globalSettings.GPS_Enabled = true
+	globalSettings.GpsdAddress = "localhost:2947"
```
cyoung: Should be fine to use a constant, this is something that will rarely (never) be changed in production.

main/managementinterface.go
```diff
@@ -218,6 +218,12 @@ func handleSettingsSetRequest(w http.ResponseWriter, r *http.Request) {
 					case "Ping_Enabled":
 						globalSettings.Ping_Enabled = val.(bool)
 					case "GPS_Enabled":
+						switch {
+						case val.(bool) == true && globalSettings.GPS_Enabled == false:
+							connectGpsd(globalSettings.GpsdAddress)
```
cyoung: We could make the simplifying assumption that gpsd should always be running and always connect, but perhaps not use the data if `GPS_Enabled` is `false`.

PR #550, opened 2016-12-17
FIX Stratux so it works again!!! a.k.a. added stratux-wifi.sh to update scripts and correct pathing issue

peepsnet:
I added the needed lined to include stratux-wifi.sh  when the update scripts is run


Ergonomicmike:
And the last two commits prior to this bug fix should be removed, since they don't work. (Per peepsnet in Slack.)

image/stratux-wifi.sh
```diff
@@ -20,7 +23,7 @@ RPI_REV=`cat /proc/cpuinfo | grep 'Revision' | awk '{print $3}' | sed 's/^1000//
 if [ "$RPI_REV" = "a01041" ] || [ "$RPI_REV" = "a21041" ] || [ "$RPI_REV" = "900092" ] || [ "$RPI_REV" = "900093" ] && [ "$EW7811Un" != '' ]; then
  # This is a RPi2B or RPi0 with Edimax USB Wifi dongle.
  DAEMON_CONF=/etc/hostapd/hostapd-edimax.conf
- DAEMON_SBIN=/etc/hostapd/hostapd-edimax
+ DAEMON_SBIN=/usr/sbin//hostapd-edimax
```
cyoung: Double slash.

cyoung:
Removed `update-stratux-v1.1b1-3e663c89c8.sh` and `update-stratux-v1.1b1-de93c79a07.sh`.

image/stratux-wifi.sh
peepsnet: Done

peepsnet:
Should be GTG now!!

cyoung:
Thanks!

PR #554, opened 2016-12-20
Edimax mac check for any RPIs (pi0/2/3)

peepsnet:
I know we are looking to move away from the PI2 but there is always the possibility of a user losing the onboard wifi by hardware failure.  

I believe this would allow for a user to add an Edimax dongle to the PI3 and still get the system to work. Also, this should work for the PI0 with an edimax dongle

NOTE: On my PIR3 I use an Edimax Dongle as well as the onboard. It is almost a 100% guarantee that the USB-based dongle is detected first, assigning it to WLAN0. I use one WLAN for the Stratux Hotspot and the other to connect to my home/cell wifi for internet access and development/updating. This allows me to access my Stratux via SSH and still have internet access. I know I can use the ETH0 for this also. This is difficult because of the case fan blocking the ETH0 plug and the case being closed most fo the time doensn;t make it any easier .

I believe the previous method of giving weight to the RPI version is not relevant here.  Since the use of wlan0 for the Access Point is hard-coded in the hostapd.conf/hostapd-edimax.conf to wlan0 we only need to check wlan0's MAC and compare to known MACs for Edimax dongles. 

I am using http://www.adminsub.net/mac-address-finder/edimax as the list of possible MAC address ranges

If the MAC at wlan0 belongs to an Edimax dongle then the edimax driver and config file must be used regardless of RPI version(RPI0/2/3).

Note: Another approach is to check the MAC at wlan0 and if it is an EDIMAX MAC then see if there is another WLAN available without an EDIMAX MAC and use it by dynamically modifying the configuration files. But if this fails... we are back to needing to use the EDIMAX drivers... SO...

That seems like much more work to get to the same place...

### DIFF
--- a/image/hostapd_manager.sh
+++ b/image/hostapd_manager.sh
@@ -270,12 +270,18 @@ sleep 0.5
 echo "Killed..."
 echo ""
 echo "ifdown wlan0..."
+echo ""
 ifdown wlan0
 sleep 0.5
+echo "ifdown done..."
+echo ""
 echo "ifup wlan0..."
-echo "Calling Stratux WiFI Start Script(stratux-wifi.sh)..."
+echo "This calls the Stratux WiFI Start Script(stratux-wifi.sh)..."
+echo ""
 ifup wlan0
 sleep 0.5
+echo ""
+echo "ifup done..."
 echo ""
 echo ""
 echo "All systems should be up and running and you should see your new SSID!"
--- a/image/stratux-wifi.sh
+++ b/image/stratux-wifi.sh
@@ -4,27 +4,37 @@
 /usr/bin/killall -9 hostapd hostapd-edimax
 /usr/sbin/service isc-dhcp-server stop
 
+#EDIMAX Mac Addresses from http://www.adminsub.net/mac-address-finder/edimax
+#for logic check all addresses must be lowercase
+edimaxMac=(80:1f:02 74:da:38 00:50:fc 00:1f:1f 00:0e:2e 00:00:b4)
+
 #Assume PI3 settings
 DAEMON_CONF=/etc/hostapd/hostapd.conf
 DAEMON_SBIN=/usr/sbin/hostapd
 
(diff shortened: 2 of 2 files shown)


Ergonomicmike:
Clever. I don't see myself plugging a spare Edimax in my Stratux if it fails in flight (would it hot plug or require a restart?) - I treat these failures like a landing gear problem. (E.g., I accept the failure & I'll deal w/ it once I'm safely on the ground.) But I like that the code doesn't need to know which version of the Pi one is using. Seems cleaner that way and better for future development. And this option allows R Pi 3 users to have a  2nd Wi-Fi, which a lot of users seem to want.

What happens if an R Pi 2 user plugs two Edimax's in?

peepsnet:
> What happens if an R-Pi 2 user plugs two Edimax's in?

The code only looks at what your RPI had loaded to WLAN0. You could, theoretically have one EDIMAX per USB port and it should still work.

>  don't see myself plugging a spare Edimax in my Stratux if it fails in flight.

I agree, However, I could see a failure in flight with a second dongle installed for redundancy, or after the failure the whole RPI would not be junk as far as Stratux is concerned.

Ergonomicmike:
The reset would have to be power down reset, right? 'Cause if you lost WiFi , can't reset via the UI. In which case I'm still kinda hosed, 'cause I've been leaving the Pi on the hat rack.

peepsnet:
> The reset would have to be power down reset, right?

afirm!


>  In which case I'm still kinda hosed, 'cause I've been leaving the Pi on the hat rack.

http://3.bp.blogspot.com/_fX9doSZqagk/SxZg8oWoC4I/AAAAAAAABhA/9UMJpBwPOng/s320/6-33-master-alt-bat.jpg

Ergonomicmike:
Anker battery powered. (As opposed to ship's power.) Not so easy to disconnect, reconnect, especially in turbulence in IMC.

peepsnet:
Lets hold off on this PR till I do some more testing, Some ideas and issues came up

PR #555, opened 2016-12-21
Wifi lag fix

peepsnet:
1) Changed the location of the code that checks and created the new hostapd file structure to execuite after ifup wlan0. Along with moving the code to stratux-wifi.sh, I changed to format of stratux-wifi.sh to work in functions so the code can stay compartmentalized for each task. 

2) I tried and tried for hours to get encryption to work with the edimax dongle. 
Since I could not get it, I removed(commented the code out so it was not lost)  the code from my hostpad_manager.sh script so it cannot be added. 

The hostapd_manager.sh script can now only change the SSID and/or channel

### DIFF
--- a/__root__stratux-pre-start.sh
+++ b/__root__stratux-pre-start.sh
@@ -15,39 +15,3 @@ if [ -e /root/update*stratux*v*.sh ] ; then
 	fi
 fi
 
-##### Script for setting up new file structure for hostapd settings
-##### Look for hostapd.user and if found do nothing.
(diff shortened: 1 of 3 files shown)


jonvadney:
Try changing the channel (i used 6) to see if that fixes encryption.    I couldn't connect to mine when I got home today even though it was working hours before.   Looks like interference from my wireless router was preventing the secure connection on the default channel 1 even though an open connection was working.

peepsnet:
ill try but i believe i tried many channels. 1,3,6

PR #556, opened 2016-12-25
Add ability to set static ip from web ui when in dev mode

jonvadney:
1) Added the ability to add a space separated list of IP's that will be sent network data for use when testing while connected to a network other than the Raspberry PI Network.  I did see that a file could be created that contained those IP's but this seems easier.  (only visible when DeveloperMode is enabled)

2) Removed Windows characters from settings.js,  (didn't want them to be inconsistent and was editing on the Raspberry PI not on Windows. 

3) Display Raw Configuration when DeveloperMode is set.


cyoung:
Well done. Thanks for submitting.

PR #566, opened 2017-02-13
UAT traffic decoding improvements

### LINKED ISSUE #565: UAT track incorrectly calculated as 0 degrees when one velocity component is zero
In discussion of #562, @Ergonomicmike noted that a (stale) UAT target was indicating a track of 0°, despite being on a 270° track.

image

This appears to be a [very old bug in traffic.go](https://github.com/cyoung/stratux/blob/9ee46170ffab46581aa8135263ecfb50264d5666/main/traffic.go#L552), where airborne track will not be calculated unless both the N-S *and* E-W velocity components are greater than zero:

```
			if ns_vel != 0 && ew_vel != 0 {
				//TODO: Track type
				track = uint16((360 + 90 - (int16(math.Atan2(float64(ns_vel), float64(ew_vel)) * 180 / math.Pi))) % 360)
			}
```

Any aircraft on a 90°, 180°, or 270° track will therefore display as tracking due north.

The solution is simple -- change the `&&` to `||`. I plan to capture this change, plus a few minor ADS-B version 2 compatibility updates to `parseDownlinkReport()`, no later than 2/13.

AvSquirrel:
Assorted improvements to UAT traffic decoding  --  `parseDownlinkReport()` in `traffic.go`:

- Calculate ground track correctly when one velocity component is zero (fixes #565)
- Correctly parses callsign and squawk code from UAT version 2 messages
- Adds parsing of emergency / priority status, and passes it to GDL90 output
- Parses additional integrity and capability characteristics (SIL, SDA, NACv, and ADS-B In flags) and writes them to stratux.log when in DEBUG (verbose log) mode

In UAT version 1 (DO-282A) the CSID bit was always set to one, and the eight callsign characters always represented the callsign. In DO-282B, these characters represent a "flight plan ID" (i.e. the Mode 3/A squawk code) if CSID is set to zero. Previous Stratux handling ignored CSID, and always decoded callsign as callsign. This caused all current / rule compliant UAT transmitters apparently alternate between displaying callsign and squawk code. The improved handling looks at CSID, and alternately decodes the callsign characters as either callsign or squawk code, as appropriate.

Parsing of additional integrity characteristics is primarily a debug tool to more easily identify noncompliant emitters for dupe debugging, and to help identify which aircraft are capable of triggering the TIS-B "hockey puck". Similar changes could be made to a future dump1090 revision.

This PR also resolves a bug that turned on sqlite logging when replaying a legacy UAT replay log, and adds an Excel spreadsheet that documents the UAT version 1 and 2 formats and allows offline decoding of UAT downlink messages.


main/traffic.go
```diff
@@ -549,7 +627,7 @@ func parseDownlinkReport(s string, signalLevel int) {
 			}
 		}
 		if ns_vel_valid && ew_vel_valid {
-			if ns_vel != 0 && ew_vel != 0 {
+			if ns_vel != 0 || ew_vel != 0 {
```
cyoung: 👍

PR #575, opened 2017-03-21
Oled display integration

jamez70:
Up to you if you'd like this in. It may need some cleaning up, but it has been stable for over a week now running


jaywelch:
To expand on this, I had an idea the other day to add an LED display to the Stratux case and have it display UTC time with small indicators should there be errors (GPS, temp, power insufficient, etc). Looks like it'd be relatively easy from what you've done already.

main/oleddisplay.go
```diff
@@ -0,0 +1,450 @@
...
+	switch blinker {
+	case 0:
+		blinkstr = "-"
+	case 1:
+		blinkstr = "\\"
+	case 2:
+		blinkstr = "|"
+	case 3:
+		blinkstr = "/"
+	case 4:
```
sfumarolo: 4-7 are the same as 0-3, so this loop can be cut in half.

Issue #580, opened 2017-03-30
GDL-90 traffic alert status


semaphore-digital:
I've been flying with stratux connected to Foreflight, and also pushing data to my Advanced Flight 4500s EFIS via GDL-90 serial.  Everything's working well and I've done a [writeup](http://www.vansairforce.com/community/showthread.php?t=148044) to share with other pilots with similar config; this is one of a few dual-band In solutions that work with this hardware, and by far the most affordable.

In the build and testing, I did have one issue that required a change & recompile of the stratux code.  Because the GDL-90 "Traffic Alert Status" byte is hard-coded as 1 ("indicates that a traffic alert is active for this target" per the GDL-90 spec), the EFIS displayed all received targets in alert status.

Changing 
`msg[1] = 0x10 | ti.Addr_type` ([1.1b1 at 248](https://github.com/cyoung/stratux/blob/0107b59010f5b816833a60c3e6b3e5ad07f6e1ec/main/traffic.go#L248)) 
to set the alert status to 0, as:
`msg[1] = 0x00 | ti.Addr_type` 
allowed the EFIS to handle the alert determination on its own; on my particular display this shows distant traffic as a hollow diamond, nearby traffic as a filled diamond, and alert traffic as a yellow circle.

I tested Foreflight 8 and a WingX trial with the status set to 0, and it didn't seem to make any difference in display there.  I can't speak to how other EFISs or EFBs handle the difference, though.

cyoung:
Thanks for doing initial testing on this. We'll have to try with a few other EFBs to see if they even pay attention to the bit.

ChiefPilot:
I've seen this same issue with certified equipment (Avidyne IFDs) and from reading the GDL90 ICD it seems this flag should only be set when the MFD is supposed to show a traffic alert.   Pull request to replace the hardcoded bit with logic that examines distance to target is inbound and conditionally sets the alert flag is inbound.

cyoung:
Needs EFB testing.

http://updates.stratux.me/builds/update-stratux-v1.2r2-9e12f69318.sh

Ergonomicmike:
@cyoung What do I look for on my EFB?

ChiefPilot:
[...]

PR #581, opened 2017-04-03
Remove hardcoded alert flag on traffic report and replace with simple…

ChiefPilot:
… logic to set it only if a target is within 2nm.   This addresses issue #580 and has been flight tested with Avidyne IFD equipement.


ChiefPilot:
Ugh.  Gooned up the merge.  New PR incoming.

PR #582, opened 2017-04-03
Remove hardcoded alert flag on traffic report and replace with simple…

ChiefPilot:
… logic to set it only if a target is within 2nm.   This addresses issue #580 .


ChiefPilot:
Cool - thanks!

ChiefPilot:
I've flown this with FF 8 and am trying FF 9 in the air with it this afternoon.   No access to wingx or other EFBs at the moment.   I'd be super curious to see how they handle this or if they even pay attention to that bit.

PR #585, opened 2017-04-21
Galileo updates and stratux-help fix

kjablonski:
Adjusted startup script to allow for UBX8 GPS modules to accept Galileo signals.  During testing I found that including Galileo signals significantly improved accuracy until SBAS takes over.
testing galileo

Also fixed a typo in stratux-help command

I'm not sure if i need to split the fixes into separate pull requests since they deal with 2 different things


main/gps.go
```diff
@@ -254,8 +254,8 @@ func initGPSSerial() bool {
 		// Byte order for UBX configuration is little endian.
 
 		// Set 10 Hz update to make gpsattitude more responsive for ublox7/8.
-		p.Write(makeUBXCFG(0x06, 0x08, 6, []byte{0x64, 0x00, 0x01, 0x00, 0x01, 0x00})) // 10 Hz
-		//p.Write(makeUBXCFG(0x06, 0x08, 6, []byte{0xc8, 0x00, 0x01, 0x00, 0x01, 0x00})) // 5 Hz
+		//p.Write(makeUBXCFG(0x06, 0x08, 6, []byte{0x64, 0x00, 0x01, 0x00, 0x01, 0x00})) // 10 Hz
+		p.Write(makeUBXCFG(0x06, 0x08, 6, []byte{0x06, 0x00, 0xF4, 0x01, 0x01, 0x00})) // 2 Hz
```
cyoung: Maybe we could detect a UBX8 receiver and make those changes only for those? Is the reason for using a slower update speed a limitation when using Galileo?

cyoung:
Thank you for submitting this. Multiple PRs isn't needed since the typo changes are so minor, but thanks for being mindful of it.

main/gps.go
kjablonski: That's a good idea, we should be able to move the logic into line 288 ish where the UBX8 logic is.  The reason for the slower update speed is the 3.01 firmware for ubx8 restricts the update speed to 3hz while using 3 GNSS systems.  I couldn't find a valid config string that gave me 3Hz so I went with 2.

I'm not sure how to make changes in the PR, so i will back it out and resubmit

cyoung:
https://github.com/cyoung/stratux/commit/7d7c2abe354256e9a7cf0afa3d37ba5e38882b68

PR #586, opened 2017-04-21
Enable Galileo and updates to stratux-help

kjablonski:
Remake of previous PR.  Updated to include cyoung suggestions


cyoung:
Ref #585.

cyoung:
The PR will automatically update when you make commits to your fork.
There are a couple things that need changing, I see that it won't compile since `updatespeed` isn't defined. Will fix it in the morning unless you get to it before I do.

PR #591, opened 2017-05-06
Support Raspberry Pi 3B with newer kernels

lukepalmer:
Building with current raspbian and HEAD, stratux will not start on a Raspberry Pi 3B due to a problem in embd that causes hardware to not be recognized.

kidoman/embd#79 gets things working, but @kidoman has been busy with a startup and has not merged PRs in a very long time.

Because go get does not know how to pull from branches, to get things working again I merged this into a fork and referenced that from within stratux. Hopefully if embd sees some love in the future we can undo this change.

### DIFF
--- a/main/gps.go
+++ b/main/gps.go
@@ -20,9 +20,10 @@ import (
 
 	"bufio"
 
-	"github.com/kidoman/embd"
-	_ "github.com/kidoman/embd/host/all"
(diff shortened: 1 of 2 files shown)


lukepalmer:
@kidoman actually just gave me push access to embd, so this will resolve itself shortly!

PR #593, opened 2017-05-11
Rework fan PWM control

lukepalmer:
I was trying to reduce power consumption for a battery-powered stratux and found that the cooling fan included with available cases accounted for up to 15% of the power budget. We can reduce that quite a lot.

The fan control script included with the stratux image is a good idea but is in an unhappy state, so I rewrote it into something that does the job it was trying to do.

Problems fixed by the rewrite include:
1. There's no need to try and cool the CPU to 40C. We're going to get into thermal trouble only around 80C, so set a better default target of 65C (which is still conservative). The fan will run much less.
1. We were trying to do PWM control via software bit-banging. Instead, use onboard hardware PWM which is much better at this job. Not only do we free up CPU cycles but we can use a far higher PWM clock  frequency. This makes the fan audibly whine a lot less plus is more efficient.
1. We were trying to scale PWM duty cycle all the way down to zero. A typical fan is going to stop spinning below around a 50% duty cycle, so at that point we're actually creating heat instead of cooling and are wasting power. Fix this by turning the fan off entirely below a useful duty cycle.

The new program also shares temperature polling code with the existing codebase.


Helno:
Nice to see someone who actually knows how to properly code this working on it. 

I chose 40C since that seems to keep everything at a reasonable temperature. I found that if  you set it higher the SDR's would get unreasonably hot before the CPU sensed it. Same reason why I recommend having the fan pull air from the case and blow it out. The main heat load is the SDR's and the CPU is only used as a temperature measuring device.

The fans I have been using work reliably down well below 30% duty cycle but will not start up at those low levels. A suggestion in this department would be to start the fan at 100% on startup and then ramp down to 50% just to give an indication that things have started up correctly.

cyoung:
@lukepalmer - thanks for sending this in.

I think the majority of people that will be using the fan controller also have the BMP280 installed. Do you guys think that using that to measure the air temperature inside the case is any better than CPU temp?

Makefile
```diff
@@ -33,6 +37,10 @@ www:
 install:
 	cp -f gen_gdl90 /usr/bin/gen_gdl90
 	chmod 755 /usr/bin/gen_gdl90
+	cp -f fancontrol /usr/bin/fancontrol
```
cyoung: Needs added to `selfupdate/makeupdate.sh` and `image/mkimg.sh`.

lukepalmer:
Gotcha, that makes more sense with cooling the SDRs. 

@helno do you have a sense of at what point the SDRs have problems due to temperature? Since it's a radio it'll have better reception when it's colder but that's hopefully pretty minor. The datasheets for the RTL chips aren't public, so I don't know what temperature range they're spec'ed for.

@cyoung that's a good idea, we could make it do either. I wouldn't mind doing a second PR with that as an option. I think I have one of those around somewhere.

main/fancontrol.go
```diff
@@ -0,0 +1,190 @@
...
+func fanControl(pwmDutyMin int, pin int, tempTarget float32) {
+	cPin := C.int(pin)
+	C.wiringPiSetup()
+	C.pwmSetMode(C.PWM_MODE_BAL)
+	C.pinMode(cPin, C.PWM_OUTPUT)
+	C.pwmSetRange(pwmDutyMax)
+	C.pwmSetClock(pwmClockDivisor)
+	C.pwmWrite(cPin, C.int(pwmDutyMin))
+	temp := float32(0.)
+	go cpuTempMonitor(func(cpuTemp float32) {
```
cyoung: IIRC there was one case previously where concurrent access to `/sys/class/thermal/thermal_zone0/temp` caused a kernel panic.

cyoung: https://github.com/raspberrypi/firmware/issues/132

Helno:
I chose 40C because it tended to result in a 50% duty cycle on a typical day resulting in a small amount of airflow over the SDR's. If it is really cold then the fan shuts down, and if it is really hot the fan will be at 100%.

The CPU rarely gets above 60C even with no fan unless it is a hot day.

What my original plan for this was to turn the fan off if the CPU dropped below 30C but it was beyond my limited ability to write code.

cyoung:
@lukepalmer - BMP280 driver in the `ahrs_dev` branch. I don't think it pulls temperature yet.

main/fancontrol.go
lukepalmer: Interesting. I'm not sure what we can do about this, other than report a firmware bug if we observe it (though the reports in the issue look to be quite old). Note that the python script in the current image is also polling the same thing, so we're not really changing anything. 

Open to ideas of course.

lukepalmer:
Thanks for the comments. Will address the issues raised and look at an alternative temperature source separately.

main/fancontrol.go
cyoung: Fair enough. Let's just keep an eye on it.

JohnOCFII:
A couple of comments.  

- I've got the fan running in two units, one with the BMP280 and one without.  Ideally code for both cases.  
- Using the current code from @Helno I changed the temp from 40C to 50C, just so it doesn't fire up the fan in the house, because it's noisy. :)

lukepalmer:
@JohnOCFII yup that makes sense. We'll start with CPU temp as a first pass. Note that in this PR the target temperature is 65C but you can override that with a command line argument "-temp", which you'd likely want to place in your systemd startup script to make permanent.

deandayton:
FWIW, I played with fancontrol.py to deal with some of these issues. Here are a few of the things I did.
1) If the algorithm called for a duty cycle under 20%, set the PWM to 20%. The fan won't continue running below 20%
2) If the algorithm goes from 0 to non-zero, run the fan at 70% for 1 sec, then run the fan per algorithm(see note 1). I couldn't reliably start the fan under around 40% duty cycle
3) If the temperature is below 15 at startup, initialize fan speed to 0. Why run the fan if it's in a cold airplane?
4) If the temperature is above 65 at startup, initialize fan speed to 100. Start cooling now, if we are in an closed airplane on a hot ramp.

lukepalmer:
@deandayton I like the idea of a startup power setting and I'll look into adding. The other cases you mention should be all covered by this change.

deandayton:
Great. I'm glad you are looking into this. Let me know if I can help.

cyoung:
@lukepalmer - I'll merge now and add to `selfupdate/makeupdate.sh` and `image/mkimg.sh` since they're easy changes. Subsequent improvements can go in another PR.

cyoung:
70F room temp, running for 20 mins. Looked like the CPU temp stabilized at 60C and didn't go over it. Pic with the fan not having run for the whole 20 min:

flir0769

Case is pretty heat soaked. This is what it looks like after starting the fan 100% and letting it run for 1.5 minutes (after taking the above pic).

flir0783

JohnOCFII:
Curious if running the fan longer continues to lower temperatures.
Also curious to know if the fan is set to run from the first time it crosses heat threshold (50C? 60C), if the case stays cooler overall, or if it eventually heats up as much as running without the fan, then running the fan for 5 minutes.

lukepalmer:
The IR pictures are great! 

A piece of data we don't have is: how hot is too hot? I'm going to try some experiments with higher temperature settings. I'm trying to get power use down fairly aggressively as part of an internal battery I'm working on.

cyoung:
Test: no fan vs fan always on with minimum settings (pwmDuty=20, pwmClockDivisor=128).

No fan. Run time 506s, 0.553 Wh => 3.93 W (dual band, low power v2, GPS, AHRS board and fan controller).

no_fan

Always on. Run time 810s, 0.955 Wh => 4.24 W.

min_fan


Here's what it looks like after about 45 minutes.


flir0789

cyoung:
v1.3b2 defaults at room temp. Run time 947, 1.107 Wh => 4.208 W.

v1 3b2

So the current cost is 0.274 W.

PR #595, opened 2017-05-13
Rework fan control, add low power automatic shutdown

lukepalmer:
When used with a power supply that can indicate low battery power, we can use a GPIO pin to cleanly shut down the OS when the battery is low. (note: article on using a boost converter with LiPo batteries inside a stratux case coming soon).

Generalize the fancontrol script and now call it hwcontrol, as all the other stuff it needs to do is nearly the same.

Add an optional flag for pwm hardware clock. Because:

@cyoung I see you changed my setting of 10kHz to 150kHz. Because PWM is only efficient if you keep the transistor in saturation or off as much of the time as possible, and the fan motor has an internal inductance, at 150kHz the switching transistor is spending a lot of time in in-between states and making heat instead of being all the way on or off. Folks can now set this value to whatever they want, but I'd recommend something lower. Original divisor was 2048 but we could even go to 1024 if you think that's too noisy.


cyoung:
What fan are you using? What settings were you able to use such that the switching wasn't audible?

cyoung:
http://www.aosmd.com/pdfs/datasheet/AO3400A.pdf is what most people are using.

Looks to me like it can switch upwards of 10 MHz before this is a real issue, what's missing?
screen shot 2017-05-13 at 6 49 32 pm

lukepalmer:
I probably have a similar FET around somewhere, let me dig one out and compare. I'm mostly using the fans that come with the summitlink cases.

Generally though because the fan is an inductive load, there is a time lag between when  the transistor turns on and when current starts flowing (that's fairly hand-wavy to keep it short). The more you raise the switching frequency the more dominant that lag becomes.

Here's a low-tech experiment to observe. You can try this with gpio from wiringPi. Set a duty cycle of around 50% and a PWM clock around 10kHz. Observe how fast the fan is spinning. Then raise the clock to whatever you like keeping the duty cycle the same. The fan will slow down. The difference represents a loss in efficiency.

lukepalmer:
Let's move the PWM frequency discussion over to #599. For the purposes of this change, the PWM frequency is now configurable with a command line switch.

lukepalmer:
@cyoung are you ok with this PR? I'd like to add more on top of it and don't want to get too far out of sync.

cyoung:
@ChiefPilot - this change doesn't change anything relevant to GPS/GDL90 that I can see. Not sure which change you're referring to.

lukepalmer:
Just did the following:

- Merged with upstream
- Added recommended defaults tuned to the SunOn fan with suggested settings for those using the pi fan

lukepalmer:
@cyoung is there anything else you'd like to see on this, or could we merge?

pc500:
I am looking for a way to hard-wire the Stratux device to plane power.  The battery powers it, and then eventually dies when the battery pack dies.  When I turn the plane back on, the battery starts charging again and Stratux automatically powers up.  This is not ideal.  I'm trying to emulate the behavior the Stratus supports for this use case.

Will the change here with power control and a specific battery pack solve my use case?

lukepalmer:
@pc500 Yes, if you used the automatic low battery shutdown it would do what you're asking. Unfortunately this PR has been sitting for quite a long time without signs of movement, so I'm not sure if this will happen.  :(

cyoung:
https://github.com/cyoung/stratux/blob/68e95193fc2e68bcc02c2266d6026bfea935bc0d/main/hwcontrol.go#L114-L130

Saving for later in case we have hardware that uses this, otherwise the other changes made their way in.

PR #596, opened 2017-05-13
Added toggle switch for Galileo tracking. Slightly reduced number Galileo tracking channels to ensure tracking performance of mature GNSS systems.

dndx:
I attempted to do a hot reinitialization of the GPS receiver by setting `globalStatus.GPS_connected` to `false` and force `initGPSSerial` to be called. But despite new configs being sent, those settings are not taking effect until the GPS receiver was power cycled (for GPS connected with GPIO, that requires to pull the power plug and re-connect). 

Anyway, after a power cycle, the settings will take effect correctly.

Leaving Galileo to OFF by default for reasons phrased inside #589.

Supersedes #589.

Tested with my GPIO RY836AI build and works great.

screen shot 2017-05-13 at 9 30 32 pm

### DIFF
--- a/main/gen_gdl90.go
+++ b/main/gen_gdl90.go
@@ -1043,6 +1043,7 @@ type settings struct {
 	WatchList            string
 	DeveloperMode        bool
 	StaticIps            []string
+	Galileo_Enabled      bool
 }
 
 type status struct {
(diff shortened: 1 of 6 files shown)


dndx:
cc @kjablonski

PR #597, opened 2017-05-14
Removed Galileo nav rate change. Slightly reduced Galileo's maximum tracking channel.

dndx:
The current setting we use is incorrect for a 2 Hz configuration. The update interval is set to 166 Hz and sample for calculation is set to 500 and is outside of the chip's capability. This causes NEO-M8 to fallback to the default 1 Hz update rate and is not really desirable.

According to [answer from Ublox forum](https://forum.u-blox.com/index.php?qa=2342&qa_1=enable-galileo-navigation-rate-need-where-find-rate-change&show=2347#a2347), we actually do not need to tune down the nav rate when enabling Galileo at all as the chip will automatically scale down as necessary. I have been able to confirm this behavior, that is, the nav rate only goes down when Galileo is being tracked and stays at 10 Hz when no Galileo satellite was used in the calculation (and appears to scale down linearly as the number of Galileo satellite tracked goes up). Adding a toggle switch for Galileo is also no longer necessary as older chip user should receive no negative impact from enabling Galileo.

This PR also reduces the maximum tracking channel of Galileo from 8 to 5 as there are currently only 11 operational satellite inside the Galileo system and it is not likely to observe more than ~5 satellites at a given point on earth. With this change we free up 3 more tracking channels for GPS/GLONASS/SBAS.

**Before:**
screen shot 2017-05-13 at 9 30 32 pm

**After:**
screen shot 2017-05-13 at 9 54 43 pm

### DIFF
--- a/main/gps.go
+++ b/main/gps.go
@@ -288,7 +288,6 @@ func initGPSSerial() bool {
 			//log.Printf("UBX8 device detected on USB, or GPS serial connection in use. Attempting GLONASS and Galelio configuration.\n")
 			glonass = []byte{0x06, 0x08, 0x0E, 0x00, 0x01, 0x00, 0x01, 0x01} // this enables GLONASS with 8-14 tracking channels
 			galileo = []byte{0x02, 0x04, 0x05, 0x00, 0x01, 0x00, 0x01, 0x01} // this enables Galileo with 4-5 tracking channels
-			updatespeed = []byte{0xF4, 0x01, 0x01, 0x00, 0x01, 0x00}         // Nav speed 2 Hz for enabling Galileo
 		}
 		cfgGnss = append(cfgGnss, gps...)
 		cfgGnss = append(cfgGnss, sbas...)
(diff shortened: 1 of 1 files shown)


VirusPilot:
I confirm that this 2 Hz configuration is wrong. No need to change the RATE configuration from what is used for ublox 7.
Furthermore as of February 2019 there are significantly more GALILEO satellites up so that the max. number could be set to 10 which is the maximum that the ublox 8 chip can do anyway.

cyoung:
Anyone able to test with higher tracking channel settings, now that there are more active Galileo satellites?

VirusPilot:
We are successfully running ublox 8 units with 10Hz and Galileo now for more than a year in the Stratux Europe. I can prepare a pull request if you want to bring these changes back.

VirusPilot:
And furthermore we have set the number of tracking channels for Galileo to 10 and it workes fine. As we speak, 22 Galileo sats are active: https://www.gsc-europa.eu/system-service-status/constellation-information

PR #598, opened 2017-05-14
Added nav rate display and Galileo nav rate change for AHRS branch.

dndx:
AHRS calculation used to be inside `calcGPSAttitude()` which is no longer being called when IMU is enabled. This PR moves the calculation code to a dedicated function to ensure nav rate can still be displayed in IMU AHRS mode to help evaluating GPS performance.

Also includes changes from #597


cyoung:
This is helpful. Merged into AHRS branch.

PR #609, opened 2017-05-30
Creation of DEB package

bandi13:
I have created a way to make DEB packages for Stratux. The makeDebPkg script does everything necessary for a fresh install. My plan for this is to try to come up with a minimal image of Linux for the rPI (like MINBIAN) and then install Stratux. I hope this may even reduce power consumption.

Additionally, I'm thinking of having the official DEB packages hosted somewhere, so when the rPI is connected to the internet via ethernet, doing an 'apt-get upgrade' would update Stratux from an official repository.

The compilation time has also been significantly improved. You can run 'make -j 4' to use all four cores. The 'go build' commands also complete faster after applying the '-i' flag so it doesn't need to recompile unchanged libraries.


bandi13:
Fixed merge conflict

PR #636, opened 2017-07-13
Added Port to calls

peepsnet:
I added code to get the protocol and port for the call so it can be used through a firewall or through https if needed... who knows...

This should be a more graceful solution because it returns the results how they were made.


peepsnet:
incorporated into PR634

PR #638, opened 2017-07-16
[docs] change snippet to 'javascript' for github markdown

drnic:
github markdown `json` format doesn't accept comments; but luckily `javascript` format does. This change makes the docs look nice in github.


cyoung:
Good call, thanks for submitting.

Issue #631, opened 2017-07-05
Sighting: Stratux drop out


jzeevi:
1. Stratux version:1.3b3
2. Stratux config:

    SDR
      - [ ] single
      - [ X] dual

    GPS
      - [X ] yes
      - [ ] no
      type:  GPYes

    AHRS
      - [ X] yes
      - [ ] no

    power source: Airplane (USB Port)

    usb cable: um, yes?

3. EFB app and version: (e.g., WingX Pro7 8.6.2)  FF, beta

    EFB platform: (e.g., iOS 9.2) 10.3.x (latest, whatever version that is)

    EFB hardware: (e.g., iPad Mini 2) iPad Mini 2

4. Description of your issue:  A sighting: while on a set of long flights (2-3 hours each) over this last weekend, I saw FF showing ADS-B towers (i.e. connected to Stratux), but NO ADS-B data. Speed went to zero, position went to zero, the FF airplane went to a blue pulsing dot, "remaining" data was missing in the navlog. In one instance, the data came back by itself within a minute or so. In another case, I rebooted the Stratux (power cycle) after 2-3 minutes. When it re-occured, I simply powered off the stratux and utilized the iPad GPS for FF data (at this point, I was 2.5 hours out from my destination. In total over three days, I flew 18+ hours).  Earlier, I had checked Stratux temperature and seen that it was around 51 degrees C. Note: in my plane (Aircoupe) the GPS has sky view through the canopy. 

If possible, enable "Replay Logs", reproduce the problem, and provide a copy of the logs in http://192.168.10.1/logs/stratux/ and http://192.168.10.1/logs/stratux.log.

Ergonomicmike:
This just happened to me today, although mine might not be a representative sample. (I've sh'd up to an Eric build, and then back down to a Chris build. ) Also, my Stratux has been on for about 4 hours in the house (AC set for 83) and is quite warm in a Chris box, with no fan and copper foil shield around to make it a little oven. Am also getting "stuttering" of the web UI, as if packets aren't making it thru or are being delayed or sent real slow. https://vid.me/bGy2d

daveholland86:
I also had mine drop out completely, no traffic no GPS but the box was still running and wifi on  v1.3b2
 c6e96b5.  I was flying my checkride so I didnt trouble shoot it but it really threw me off my game (Instrument check), still passed though.  I have since updated to the new b3 but have not flown since.

jpuetthoff:
[...]

PR #660, opened 2017-08-29
isSleeping() ARP Check Addition and ffMonitor() Updates For Client Broadcasts

pmaag:
Fixes for issue #631 

isSleeping() has been updated to check the PI's ARP table to help determine if a client is sleeping or awake.  This should help clients that inconsistently reply to ICMP ping requests.

ffMonitor() has been re-activated and modified to update a client's state when the client sends a Foreflight heartbeat packet to broadcast.

Logging messages have been added to help troubleshooting.  These messages are controlled via the Verbose Message Logging toggle within the GUI.

### DIFF
--- a/main/network.go
+++ b/main/network.go
@@ -625,14 +625,13 @@ func networkStatsCounter() {
 
 /*
 	ffMonitor().
-		Watches for "i-want-to-play-ffm-udp", "i-can-play-ffm-udp", and "i-cannot-play-ffm-udp" UDP messages broadcasted on
-		 port 50113. Tags the client, issues a warning, and disables AHRS GDL90 output.
-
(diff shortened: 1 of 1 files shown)


Ergonomicmike:
Will this affect the client sleep stuff?

pmaag:
@Ergonomicmike  - I do not believe there will be any difference in behavior between this change and the existing code for sleeping clients.  Definitely would appreciate a peer review/testing confirmation of that though.

pmaag:
Just a heads-up, flew about 10 hours these past few weeks and the ADS-B Disconnected messages are still occurring even with this patch, however much less frequently. I would say I have gotten 3 messages in the past 10 flying hours. I have not had time to pull the logs, but will do so and report back.

cyoung:
I wonder how this will do on iOS 11?

Issue #705, opened 2018-04-03
gps.go, questionable handling of mySituation.AHRSGyroHeading in func makeAHRSGDL90Report()


d-hoke:
1. Stratux version: v1.4r4 (5af6d77)
2. Stratux config: irrelevant (I think) but its dual radio, GPS/AHRS, RPi3 

4. Description of your issue: 
Following code from main/gps.go:func makeAHRSGDL90Report():
`		if isAHRSInvalidValue(mySituation.AHRSGyroHeading) {
			hdg = roundToInt16(mySituation.AHRSGyroHeading * 10) // TODO westphae: switch to AHRSMagHeading?
		} else {
			hdg = roundToInt16(float64(mySituation.GPSTrueCourse))
		}
`

seems questionable in at least two respects:
1) Appears to be using a value that is found to be invalid
2) in the else branch, the value used is not scaled by 10, as seems to be done with all the other values in that general section of code subject to 
`if isAHRSValid() {
  ...
}
`

cyoung:
Not able to investigate now -- @westphae, any thoughts?

westphae:
I believe this comment is correct.  Working on validating the result in the gdl90 stream; once validated, will correct, test and push.  Currently this has little to no impact as no EFB's that I know of report the gyro heading.

westphae:
Created a PR at https://github.com/cyoung/stratux/pull/720 to fix this.

cyoung:
Excellent.

PR #720, opened 2018-05-29
Fix gdl90 AHRSGyroHeading reporting.

westphae:
Minor fix to address https://github.com/cyoung/stratux/issues/705.  I don't know of any EFB's that use AHRSGyroHeading, so validated that behavior is now as expected using a gdl90 udp stream listener.


cyoung:
LGTM. Would like to move away from the magic 'invalid' values, though. Especially since implementing new AHRS messages (like the FF AHRS message).

PR #668, opened 2017-09-22
Merge ahrs_dev branch into master.

cyoung:

PR #734, opened 2018-08-06
Wifi Configuration on Settings Page

westphae:
Allow user to configure wifi through the web UI: change SSID, use WPA and set passphrase, change wifi channel.


westphae:
This is a clean merge of the code in https://github.com/cyoung/stratux/pull/668.  Primary difference from that PR is I reworked web/plates/settings.html.

cyoung:
Thanks!

ninelima:
I can get a batches of 100x 1090Mhz antenna's like this one made up for 
about $10 a pop.
Would you be interested in selling them on your website?

I've test flown the one pictures a few times now one and it is (insofar 
as I could tell, without a S-Parameter test set)  just as good as the 
thick one's 4 times as long. The range is (again subjectively) exactly 
the same. In terms of convenience the little dinky ones are much nicer, 
since they fit better on the dash.


Cheers
Davie

PR #748, opened 2018-09-08
Allow a nonstandard port to be used

peepsnet:
This can allow for the info to be retrieved from behind a firewall with different ports. 

Whichever port the script is called from it will use to make calls back

### DIFF
--- a/web/js/main.js
+++ b/web/js/main.js
@@ -1,29 +1,27 @@
 // application constants
-var URL_HOST_BASE            = window.location.hostname;
-var URL_SETTINGS_GET         = "http://" + URL_HOST_BASE + "/getSettings";
-var URL_SETTINGS_SET         = "http://" + URL_HOST_BASE + "/setSettings";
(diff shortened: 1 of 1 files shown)


JCBird1012:
Why would this be necessary considering Stratux's use-case? Stratux's info page should mainly be accessed from a mobile device connected directly to the Stratux's wifi network...
Is the implication that there will be a firewall between your EFB device of choice and Stratux itself — that is a very non-standard setup and something that either @Helno or @cyoung should weigh in on further.

peepsnet:
More compatible with any situation that one may be found within. 

There was a request to force the data to IP/IPs (See Settings -> Static IPs) so a user could hook the Stratux to the WIF in the aircraft and send data to the tablets the Pilots use. 

It is then reasonable to believe many different options can be used. Why not give maximum compatibility to the user. What does it hurt??

It can only help!

peepsnet:
upon further review...

It appears the changes are in the current master...
I do not understand it...
https://github.com/cyoung/stratux/blob/master/web/js/main.js

PR #749, opened 2018-09-24
GDL90 GPS Altitude msl vs WGS-84 ellipsoid

PepperJo:
This pull request introduces a switch to the settings page to switch between GPS altitude MSL or height above WGS-84 ellipsoid for the GDL90 ownship geometric altitude message. GDL90 specifies ownship geometric altitude as height above WGS-84 ellipsoid but some EFBs expect MSL altitude. We leave the default to MSL but allow switching to height above ellipsoid.

### DIFF
--- a/main/gen_gdl90.go
+++ b/main/gen_gdl90.go
@@ -441,9 +441,16 @@ func makeOwnshipGeometricAltitudeReport() bool {
 	msg := make([]byte, 5)
 	// See p.28.
 	msg[0] = 0x0B                                // Message type "Ownship Geo Alt".
-	alt := int16(mySituation.GPSAltitudeMSL / 5) // GPS Altitude, encoded to 16-bit int using 5-foot resolution
-	msg[1] = byte(alt >> 8)                      // Altitude.
-	msg[2] = byte(alt & 0x00FF)                  // Altitude.
+
+	var GPSalt float32
+	if globalSettings.GDL90MSLAlt_Enabled {
+		GPSalt = mySituation.GPSAltitudeMSL
+	} else {
+		GPSalt = mySituation.GPSHeightAboveEllipsoid
+	}
+	encodedAlt := int16(GPSalt / 5)    // GPS Altitude, encoded to 16-bit int using 5-foot resolution
+	msg[1] = byte(encodedAlt >> 8)     // Altitude.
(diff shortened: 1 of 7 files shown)


cyoung:
@PepperJo - *all* of the EFBs on our compatibility list expect the altitude to be MSL. It was implemented wrong by most ADS-B portables, so we deviated from the spec on this since MSL was the "de-facto" standard. What EFBs expect it to be WGS-84, and is it possible to auto-detect them so that we can avoid adding extra options?

PepperJo:
From this forum posts: http://forums.skydemon.aero/Topic19950-11.aspx
Skydemon expects WGS-84. Autodetection might be possible but what e.g. multiple different EFBs are connected?

cyoung:
Stratux tracks client states. We used to shut off AHRS output only for detected ForeFlight clients because of a bug in ForeFlight. Something similar can be done to output WGS-84 if SkyDemon can be detected.

PepperJo:
I will take a look and update the pull request accordingly if I find a decent solution.

cyoung:
See `networkConnection` in `main/network.go`, this is where connection-level states are stored. We could make another capability flag to extend `NETWORK_GDL90_STANDARD`.

VirusPilot:
> @PepperJo - _all_ of the EFBs on our compatibility list expect the altitude to be MSL. It was implemented wrong by most ADS-B portables, so we deviated from the spec on this since MSL was the "de-facto" standard. What EFBs expect it to be WGS-84, and is it possible to auto-detect them so that we can avoid adding extra options?

@cyoung - your EFB compatibility list is missing some **very** popular European EFBs which I already tested successfully with Stratux:

SkyDemon (only Traffic):
- Stratux Settings: GDL90 MSL Altitude: **off**

Sky-Map (Traffic & AHRS):
- Setup - IP Adresse: UDP, Port: 4000
- Stratux Settings: GDL90 MSL Altitude: on

Air Nav Pro (only Traffic):
- Stratux Config: https://github.com/0sander/stratux_air_nav_pro
- Stratux Settings: GDL90 MSL Altitude: on

EasyVFR (Traffic & AHRS):
- Stratux Settings: GDL90 MSL Altitude: on

Airmate (only Traffic):
- Stratux Settings: GDL90 MSL Altitude: (not determined yet)

PepperJo:
I will split this pull request since it tackles 3 different problems...

PR #755, opened 2018-10-24
GDL90: traffic report convert GNSS altitude

PepperJo:
GPL90 traffic reports expect barometric altitude. Convert
GNSS altitudes in traffic information to barometric altitude.
Estimate barometric altitude from reported GPS geoid height,
own geoid separation and barometric pressure.

Signed-off-by: PepperJo <pepperjo@japf.ch>

### DIFF
--- a/main/traffic.go
+++ b/main/traffic.go
@@ -315,15 +315,24 @@ func makeTrafficReportMsg(ti TrafficInfo) []byte {
 	//
 	// Algo example at: https://play.golang.org/p/VXCckSdsvT
 	//
-	var alt int16
-	if ti.Alt < -1000 || ti.Alt > 101350 {
-		alt = 0x0FFF
+	// GDL90 expects barometric altitude in traffic reports
+	var baroAlt int32
+	if ti.AltIsGNSS {
+		// Convert from GPS geoid height to barometric altitude
+		baroAlt = ti.Alt - int32(mySituation.GPSGeoidSep)
+		baroAlt = baroAlt - int32(mySituation.GPSAltitudeMSL) + int32(mySituation.BaroPressureAltitude)
+	} else {
+		baroAlt = ti.Alt
+	}
+	var encodedAlt int16
+	if baroAlt < -1000 || baroAlt > 101350 {
+		encodedAlt = 0x0FFF
 	} else {
 		// output guaranteed to be between 0x0000 and 0x0FFE
-		alt = int16((ti.Alt / 25) + 40)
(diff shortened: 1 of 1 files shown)


PepperJo:
ping

main/traffic.go
```diff
@@ -315,15 +315,24 @@ func makeTrafficReportMsg(ti TrafficInfo) []byte {
...
 	//
-	var alt int16
-	if ti.Alt < -1000 || ti.Alt > 101350 {
-		alt = 0x0FFF
+	// GDL90 expects barometric altitude in traffic reports
+	var baroAlt int32
+	if ti.AltIsGNSS {
+		// Convert from GPS geoid height to barometric altitude
+		baroAlt = ti.Alt - int32(mySituation.GPSGeoidSep)
+		baroAlt = baroAlt - int32(mySituation.GPSAltitudeMSL) + int32(mySituation.BaroPressureAltitude)
```
cyoung: `baroAlt += -mySituation.GPSGeoidSep - mySituation.GPSAltitudeMSL + mySituation.BaroPressureAltitude`

simplifies to

`baroAlt += -mySituation.GPSHeightAboveEllipsoid + mySituation.BaroPressureAltitude`

Seems like an interesting estimation to convert HAE to pressure altitude using local data.

One thing that is missed here is how data is passed to stratux from dump1090:

https://github.com/stratux/dump1090/blob/master/net_io.c#L569-L582

HAE delta is received and passed over as `GnssDiffFromBaroAlt`.

I think this PR could be rolled up into a one line change on stratux/dump1090:

https://github.com/stratux/dump1090/blob/13d61a9a3276de26322b88c99a7426a5625f592b/net_io.c#L575

Removing the `Modes.use_hae` condition (seems unexpected to only pass the GNSS-derived altitude when the delta is available).

Have you seen many aircraft that come through with `AltIsGNSS` being true?

PepperJo: Haven't seen any ADS-B traffic where this is set however the FLARM forks use this since FLARM reports GPS altitude. The flag was already there but was not handled. I propose we either handle this in the FLARM code and remove the flag and make clear that altitude is barometric here or we convert it.

PepperJo:
What's the verdict on this pull request? Do we want it or do we rely on every module feeding data to do their own conversion? If so we should at least clarify that the AltIsGNSS does not do any conversion but sends the data as is via GDL90 which expects baro alt. Or remove the flag as suggested above.

Issue #203, opened 2016-01-13
Perfromance Anlysis


Axtel4:
Chris, if you don't feel this belongs here you may take it down.

I, like many have been curious about the performance of my Stratux unit and as such I wanted to quantify its performance.

My Stratux comprises of the R-Pi 2 B, a RY835 GPS / AHRS, 2 of the original T2  RTL-SDR dongles and various antenna configurations used during testing.  The pictures of my Statux build are available here: http://imgur.com/a/G085R

The antenna configurations will be discussed later.  The test method was to select an antenna configuration, capture the Stratux logs and post process and analyze the data.

Test Configuration Number 1:
The first test configuration consisted of a tuned Diopole Vertical antenna (one at 1090 MHz and one at 978 MHz) connected directly to the SMA connectors on the Startux unit.  Stratux was place on the dash board of my truck and was parked in a field behind the hose with a clear view to the sky to the north.  The cab and body of the truck shielded the Stratux reception to the south.  A 15 minute recording was taken and 1090ES data was post processed.  The data consisted of anywhere from 1,500 to over 5000 1090ES hits.  The range to the detected traffic was calculated using the Haversine Formula with my truck location at the center.  A Track Plot was created to visualize the range and bearing of the traffic. These track plots utilize the same methods to analyze TCAS performance data. This is a Track plot of the first configuration. 
 
![1090 whip](https://cloud.githubusercontent.com/assets/15059010/12284609/957fcaae-b976-11e5-94a9-bb209916273f.jpg)

As can been seen in this plot, the range approaches 170 nm.  This is not bad for an off the shelf RF dongle.  Now I had a baseline to use for the rest of the testing.  (It appears the range is limited due to the view angle of the traffic from my location.  I'll add more about the view angle at a future date.)  

Test Configuration Number 2:
This configuration removed both tuned antennas and feed both dongle via the tuned 1090 MHz antenna and an SMA "T".  Conventional wisdom would say that there would be a 3 dB reduction on power to each dongle and thus a reduction in range by at least one eighth.  This is otherwise know as the invers square law: Where if you double the distance, the power at the receiving antenna is one forth the previous power lever.  Let's see how the looks in practice.

![1090 whip t](https://cloud.githubusercontent.com/assets/15059010/12284748/f72f78ca-b977-11e5-967a-3c12ab9ebb97.jpg)

The above plot still shows a range out to 170 nm.  Based on this data, we still haven't lost reception of the traffic before is moves out of our view.

[...]

PR #827, opened 2020-12-06
FLARM NMEA protocol serie

claudius310:
Hello,
I installed version V1.6r1-eu021 with FLARM but I cannot use it because it does not have the possibility of outputting in RS232 according to the FLARM NMEA protocol to use all FLARM receivers.
Would it be possible because I am a perfect computer beginner to realize it because it was developed in previous versions in the US?
Best regards

### DIFF
--- a/Case designs/Readme.md
+++ b/Case designs/Readme.md
@@ -1,0 +1,2 @@
+Here are all the 3d printable cases I designed for Stratux over the years. 
+These are the editable files created in Solidworks.


b3nn0:
You are completely wrong here.
1. This shouldn't be a Pull Request but an Issue
2. This is Stratux US project, not Stratux EU. If your issue is related to Stratux EU, create the issue at b3nn0/stratux, not cyoung/stratux
3. Flarm-NMEA via RS232 works just fine if done correctly, see https://github.com/b3nn0/stratux/wiki/Stratux-Serial-output-for-EFIS's-that-support-GDL90-or-Flarm-NMEA-over-serial

main/sdr.go
```diff
@@ -62,7 +62,7 @@ type Dump1090TermMessage struct {
 func (e *ES) read() {
 	defer e.wg.Done()
 	log.Println("Entered ES read() ...")
-	cmd := exec.Command(STRATUX_HOME + "/bin/dump1090", "--oversample", "--net-stratux-port", "30006",  "--net", "--device-index", strconv.Itoa(e.indexID), "--ppm", strconv.Itoa(e.ppm))
+	cmd := exec.Command(STRATUX_HOME + "/bin/dump1090", "--gain", "37.2", "--net-stratux-port", "30006",  "--net", "--device-index", strconv.Itoa(e.indexID), "--ppm", strconv.Itoa(e.ppm))
```
cyoung, line 65: Have you tested this? I did a while ago with different antenna configurations and found changing the gain levels to be ineffective, the frontend on the SDR is overloaded before the gain levels make a difference. Beyond 100 ft with a high gain antenna and the highest gain level didn't cause reception issues.

b3nn0: We had reports of 1090ES planes getting lost when coming too close and @VirusPilot did some tests today with his 250W transponder and yes, it did indeed make quite a difference for receiving close signals.
With dump1090's dev branch (which now has optional adaptive gain), this can be tested quite nicely, as it will tune down gain to receive close aircraft - and there is indeed a strong correlation between distance to the transponder and the gain level that dump1090 determines to be appropriate to properly decode it.

Considering the variance in SDR quality/sensitivity and antennas, maybe the SDR you tested with was just one of the "more deaf" ones?

cyoung: "Deaf" wouldn't cause that behavior, I'm more referencing gain compression.

https://github.com/cyoung/stratux/issues/203#issuecomment-214565215

I can't get my own transponder signal in my aircraft using any R820T(2) receivers (I've tested them all), so I'd guess the new adaptive gain on dump1090 ends up pulling the gain back pretty far permanently when used inside a small aircraft with a transponder installed.

b3nn0: Yes, own transponder is certainly another level, VirusPilot tried on the ground, testing how far he needs to walk away to receive it. And yes, the adaptive gain is not usuable if you carry a transponder very close - but it is usuable for testing if reducing gain helps at all (and what gain works at what distance). @VirusPilot measured something like this:
- With 50db gain reception fails below 250m
- With 37.5db gain below 150m
- With 30db gain below 100m

Since 100m is certainly too low to react anyway, and we didn't want to break anything for people with minimalistic internal antennas, we kinda settled with "something around 35" by pure guessing. Not too scientific admittedly..

Issue #826, opened 2020-12-06
Makefile does not properly rebuild fancontrol on fresh build


ethanbrodsky:
1. Stratux version:  master as of commit 7eeb32acf6fa31dc86ac94f95975e573f2619ae4
2. Stratux config: N/A
3. EFB app and version N/A
4. Description of your issue. N/A

In the Makefile, this line:
	go build $(BUILDINFO_STATIC) -p 4 main/fancontrol.go main/equations.go main/cputemp.go
Builds fancontrol into a binary called "equations", for reasons I do not understand.
This causes "make install" to fail, since "./fancontrol" does not exist.

The fix is to change this line to:
        go build $(BUILDINFO_STATIC) -o fancontrol -p 4 main/fancontrol.go main/equations.go main/cputemp.go

I have this version of go installed:
root@raspberrypi:~/testcode/original/stratux# go version
go version go1.14.2 linux/arm

I'm not sure if this is an issue with using a newer version of go, but it's been something I've had to fix in order to get the base repo to build.

PR #828, opened 2020-12-08
Add pressure altitude offset setting

k3it:
A new setting allows adjustment of the pressure altitude if the sensor is not reporting it correctly (manufacturing variance?).  In the Settings web UI simply add a positive or negative offset in feet to match the aircraft's pressure altitude (i.e. the altimeter reading when set to 29.92 inHg)

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ dump978/uat2text
 gen_gdl90
 libdump978.so
 fancontrol
+equations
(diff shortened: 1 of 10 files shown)


k3it:
this also fixes #826 with the makefile change by @ethanbrodsky

6639835 (approved):

PR #895, opened 2026-07-29
README: add DroneScan Pro to the app support list

Arifur4321:


Arifur4321:
DroneScan Pro is an Android app that reads GDL90 traffic from Stratux
over UDP 4000 and plots it alongside drone Remote ID broadcasts on a
single map.

Disclosure: I'm the developer. It's free on Play:
[Play Store link ](https://play.google.com/store/apps/details?id=pro.dronescan.app)

Tested against Stratux .

Happy to reword or close this if additions to that list aren't wanted.
I notice this branch is marked deprecated — if b3nn0/stratux or
stratux/stratux is the right place for this instead, just say and I'll
move it there.

